PR #96, opened 2022-05-05
delete the stopped container by default at tear down

slice4e:
After testing with : 
redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --preserve_temporary_client_dirs --client_aggregated_results_folder ./test1

I found that I quickly ran out of 100GB of space. After some investigation, I could see that all the space is occupied by a bunch of stopped docker containers. These containers did not seem to be reused on a subsequent run I started, but rather the list grew. 

Thus, I suggest removing the unused containers automatically upon tear down.


filipecosta90 (approved):
Thank you @slice4e !

PR #98, opened 2022-05-06
Fixed 4 test-cases, which were completing too quickly 

slice4e:
Fixed 4 test-cases, which were completing too quickly


filipecosta90:
Thank you @slice4e !

PR #142, opened 2022-11-24
Fix bug 87

### LINKED ISSUE #87: [BUG]: redis-benchmarks-spec-sc-coordinator is not removing temporary folders after un
example of temporary folders on m6i machine

```
ls -lha tmp* | wc -l
14003
```

slice4e:
Fixes #87 

It seems that we were deleting "temporary_dir" but not "temporary_dir_client" 
This should be fixed now ..


filipecosta90:
@slice4e please format it:
```
1 file would be reformatted, 38 files would be left unchanged.
 ```
 you can do it via:
 ```
poetry run black . 
```

filipecosta90 (changes requested):
need to reformat using black.

slice4e:
> @slice4e please format it:
> 
> ```
> 1 file would be reformatted, 38 files would be left unchanged.
> ```
> 
> you can do it via:
> 
> ```
> poetry run black . 
> ```

Done

PR #167, opened 2022-11-30
Fix issue 130

slice4e:
The perf profiles were not being deleted and after a few days, they can end up occupying 100s of GBs of space, eventually cloggin up the machines. 

After the profiles are uploaded to S3, they are now deleted.


filipecosta90 (approved):

filipecosta90:
thank you @slice4e ! Merging.

Issue #105, opened 2022-05-13
[BUG]: Fail after memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values.yml


markovamaria:
**Describe the bug**
Tests behavior is  broken after run of test "memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values.yml":
1st run returns results, 2nd failed, 3rd run kills server (and returns strange 0 results for Ops/sec).

**To Reproduce**
Run server:
```
$ ./src/redis-server --port 6379 --dir ~/redis_server_logs --logfile fail.log  --dbfilename fail.rdb
```

In parallel run test "memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values.yml" twice, second run is failing + empty tmp* folder is appeared. 
```
$ redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --client_aggregated_results_folder ./test1 --test memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values.yml
...
# Results for memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values test-case on oss-standalone topology
|                 Metric JSON Path                 |Metric Value|
|--------------------------------------------------|-----------:|
|"ALL STATS".Totals."Ops/sec"                      |   92584.090|
|"ALL STATS".Totals."Latency"                      |       2.154|
|"ALL STATS".Totals."Misses/sec"                   |       0.000|
|"ALL STATS".Totals."Percentile Latencies"."p50.00"|       2.127|
# Results for entire test-suite
|                          Test Name                           |                 Metric JSON Path                 |Metric Value|
|--------------------------------------------------------------|--------------------------------------------------|-----------:|
|memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values|"ALL STATS".Totals."Ops/sec"                      |   92584.090|
|memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values|"ALL STATS".Totals."Latency"                      |       2.154|
|memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values|"ALL STATS".Totals."Misses/sec"                   |       0.000|
|memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values|"ALL STATS".Totals."Percentile Latencies"."p50.00"|       2.127
...

[...]

Issue #104, opened 2022-05-13
[BUG]: Fail on memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values-pipeline-10.yml


markovamaria:
Fail on test memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values-pipeline-10.yml, started from v30, still exist in [latest head](https://github.com/redis/redis-benchmarks-specification/commit/9bd0473cb2b569d84c51f0422e2322587a5b0e62) 
**To Reproduce**
```
$ redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --client_aggregated_results_folder ./test --test memtier_benchmark-10Mkeys-load-hash-5-fields-with-1000B-values-pipeline-10.yml
```
**Expected behavior**
Successful run with results

**Screenshots/CLI snippets**
Test Kills server:
```
$ ./src/redis-server --port 6379 --dir ~/logs --logfile fail.log --dbfilename my_db.rdb
Killed
```
Error message:
```
2022-05-13 17:37:32 CRITICAL Some unexpected exception was caught during local work. Failing test....
2022-05-13 17:37:32 CRITICAL <class 'redis.exceptions.ConnectionError'>
2022-05-13 17:37:32 INFO Tearing down setup

------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/Work/exp_flush/env/lib/python3.8/site-packages/redis/connection.py", line 1398, in get_connection
    if connection.can_read():
  File "/home/ubuntu/Work/exp_flush/env/lib/python3.8/site-packages/redis/connection.py", line 812, in can_read
    return self._parser.can_read(timeout)
  File "/home/ubuntu/Work/exp_flush/env/lib/python3.8/site-packages/redis/connection.py", line 317, in can_read
    return self._buffer and self._buffer.can_read(timeout)
  File "/home/ubuntu/Work/exp_flush/env/lib/python3.8/site-packages/redis/connection.py", line 224, in can_read
    return bool(self.length) or self._read_from_socket(
  File "/home/ubuntu/Work/exp_flush/env/lib/python3.8/site-packages/redis/connection.py", line 197, in _read_from_socket
    raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR)
redis.exceptions.ConnectionError: Connection closed by server.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #175, opened 2022-12-09
Added execution and updated installation section to Readme

markovamaria:
Related-to: 
https://github.com/redis/redis-benchmarks-specification/issues/105
https://github.com/redis/redis-benchmarks-specification/issues/104

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -145,13 +145,10 @@ sudo apt install docker.io -y
 
 # install supervisord 
 sudo apt install supervisor -y
-
-# install benchmark specs
-python3 -m pip install redis-benchmarks-specification --ignore-installed PyYAML
 ```
 
 
-##### Installing Redis benchmarks specification implementations
+##### Installing Redis benchmarks specification
 
 To have access to the latest SPEC and Tooling impletamtion you only need to install one python package.<br />
 Installation is done using pip, the package installer for Python, in the following manner:
--- a/Readme.md
+++ b/Readme.md
@@ -24,10 +24,11 @@ redis-benchmarks-specificaiton should be run together with redis-server in the s
 ./src/redis-server --port 6379 --dir logs --logfile server.log --save "" [--daemonize yes]
 
 # Run benchmark
-redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --client_aggregated_results_folder ./test1 --test memtier_benchmark-1key-list-100-elements-lrange-all-elements.yml
+redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --client_aggregated_results_folder ./test
 ```
 
-Option --daemonize yes allows to run redis-server in background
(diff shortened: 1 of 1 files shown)


markovamaria:
Hi @filipecosta90, could you please take a look? I've added execution section and slightly updated installation section.

By the way, I see 2 installation sections - in start of readme and below (which I updated). Should they be combined to 1?

slice4e (commented):
Maria, you are right that there are 2 "Installation" section. I would agree that we should probably keep 1 of them. I would say the second one. 
Then , notice that there are 2 ways to execute this. We have the coordinator and we have the standalone runner. 
I think that you could add your execution section before the section: Testing out redis-benchmarks-spec-sc-coordinator
You could name it: Testing out the redis-benchmarks-spec-runner  ... 
And explain the difference.

markovamaria:
> Maria, you are right that there are 2 "Installation" section. I would agree that we should probably keep 1 of them. I would say the second one. Then , notice that there are 2 ways to execute this. We have the coordinator and we have the standalone runner. I think that you could add your execution section before the section: Testing out redis-benchmarks-spec-sc-coordinator You could name it: Testing out the redis-benchmarks-spec-runner ... And explain the difference.

Hi @slice4e, I removed one extra Installation section. I moved Execution to Contribution part. As it has installation dependencies, I moved Installation section to Contribution part too. 

Honestly I slightly confused that Installation and Execution (testing out) is presented in Contribution part only.  What if I don't want to add new benchmarks, but want to see Redis pass rate on benchmarks-specification? O_o 
I would suggest to move it or duplicate before "Contributing guidelines". Please let me know what is preferable.

Also I added Navigation in top of page for easy usage.

markovamaria:
Hi @slice4e , @filipecosta90 , could you please look at CI falures? Seems like there is some connection issue, running tox from Readme locally gives me the same problem. How may I fix it?

Readme.md
```diff
@@ -5,16 +5,30 @@
...
+- [Directory layout](#directory-layout)
+  - [Specifications](#specifications)
+  - [Spec tool implementations](#spec-tool-implementations)
+- [Contributing guidelines](#contributing-guidelines)
+  - [Joining the performance initiative and adding a continuous benchmark platform](#joining-the-performance-initiative-and-adding-a-continuous-benchmark-platform)
+    - [Joining the performance initiative](#joining-the-performance-initiative)
+    - [Adding a continuous benchmark platform](#adding-a-continuous-benchmark-platform)
+      - [Installing package requirements](#installing-package-requirements)
+      - [Installing Redis benchmarks specification](#installing-redis-benchmarks-specification-implementations)
+      - [Testing out the redis-benchmarks-spec-runner](#testing-out-the-redis-benchmarks-spec-runner)
```
filipecosta90, line 22: rename it to> redis-benchmarks-spec-client-runner

filipecosta90 (dismissed):

filipecosta90:
Thank you @markovamaria . Approving!

PR #198, opened 2023-01-19
Added SADD test variations discussed in redis/pull/11595 

filipecosta90:
As being discussed in https://github.com/redis/redis/pull/11595#issuecomment-1396698804

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.1.66"
+version = "0.1.67"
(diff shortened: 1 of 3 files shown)


slice4e (approved):

zuiderkwast (commented):
Nice.

Although this benchmark is perhaps only useful for one PR. Every optimization needs its own special benchmark to prove it. What's your thoughts about that?

filipecosta90:
> Nice.
> 
> Although this benchmark is perhaps only useful for one PR. Every optimization needs its own special benchmark to prove it. What's your thoughts about that?



> Nice.
> 
> Although this benchmark is perhaps only useful for one PR. Every optimization needs its own special benchmark to prove it. What's your thoughts about that?

@zuiderkwast IMO it will still be useful to track SADD performance on a general benchmark ( we don't have specific sadd benchmarks on this spec yet ). So, lets add it :)

PR #203, opened 2023-01-27
Test runs req memory

slice4e:
If we have a test, which loads data into Redis, running this test repeatedly using "override_test_runs" may result in consuming too much memory -in the worst case crashing the redis server. 

With this pull request, we calculate the required memory per test by taking into account the number of test runs. 
The current pull request is conservative, since some tests do not grow their memory footprint over time - but we nevertheless multiply their requirements by the number of test runs.


filipecosta90:
TODO: check which tests are growing on multiple runs.

Issue #209, opened 2023-02-08
[BUG]: Failure of "redis-benchmarks-spec-cli" for retriggering of CI


markovamaria:
**Description**
Failure of "redis-benchmarks-spec-cli" for retriggering of CI in version 0.1.67

**To Reproduce**
$ redis-benchmarks-spec-cli --use-branch --from-date 2023-02-03 --redis_port 12010 --redis_host benchmarks.redislabs.com --redis_pass ***

**Screenshots/CLI snippets**
![image](https://user-images.githubusercontent.com/62143352/217573217-18ff8d9a-88eb-494f-add6-a832178c0ab5.png)

**Environment (please complete the following information):**
 - OS: Ubuntu

PR #211, opened 2023-02-08
Update timestamp conversion in client

markovamaria:
Reverting of str converting for timestamp to avoid error 'str object has no attribute timestamp'

Resolves: #209

### DIFF
--- a/redis_benchmarks_specification/__cli__/cli.py
+++ b/redis_benchmarks_specification/__cli__/cli.py
@@ -97,7 +97,7 @@ def trigger_tests_cli_command_logic(args, project_name, project_version):
     total_commits = 0
     if args.use_branch:
         for commit in repo.iter_commits():
-            commit_datetime = str(commit.committed_datetime)
+            commit_datetime = commit.committed_datetime
             if (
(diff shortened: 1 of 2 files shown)


markovamaria:
Hi @filipecosta90 , @slice4e , could you please review?

filipecosta90:
> Hi @filipecosta90 , @slice4e , could you please review?

@markovamaria please include a unit test that confirms we're OK with this change.

filipecosta90 (changes requested):
requires test. I can add it if required.

markovamaria:
> requires test. I can add it if required.

added

markovamaria:
REF 1: before fix:
image

REF 2:  after test added, before fix added: 
image

EXP: after test and fix added:
image

markovamaria:
@filipecosta90 , please take a look. I've added test and related screenshots.

slice4e (approved):

filipecosta90 (approved):

filipecosta90:
Nice. Thank you @markovamaria !

PR #232, opened 2023-04-25
fix logging level

slice4e:


filipecosta90 (approved):
Thank you @slice4e! Approving

PR #234, opened 2023-05-11
the command hincrby was populating more data 10 M , instead of 1M.

slice4e:


filipecosta90 (approved):

filipecosta90:
Thank you @slice4e. merging

PR #238, opened 2023-06-21
added filtering by platform

slice4e:
Filipe, 

Added filtering by platform, however not sure how to test it, since it depends on the builder... 

Could you please take a look? 

thanks
Martin


filipecosta90 (approved):
LGTM. Thank you @slice4e

PR #241, opened 2023-06-23
(in progress) Attempting to simplify and cleanup code. 

slice4e:
Filipe, 

I wanted to simplify the code a bit. Lets discuss this next time? 
Some functions were way too long and hard to efficiently follow. 
I made main() and blocking_read() much more manageable. And I would like to do a bit more in the same direction with the rest of the code if you agree. 
Long term, after simplifying enough, ideally I think we should get rid of the stand_alone_runner and just have an option for the self_contained_coordinator to do that. Currently they are just sharing 80% of the code ... 

Lets discuss it.

### DIFF
--- a/redis_benchmarks_specification/__self_contained_coordinator__/self_contained_coordinator.py
+++ b/redis_benchmarks_specification/__self_contained_coordinator__/self_contained_coordinator.py
@@ -122,24 +122,32 @@ def main():
     kwargs = process_args(args)
 
     logging.info("Connecting to event stream server")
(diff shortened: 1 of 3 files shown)


fcostaoliveira:
Thanks for the cleanup effort here. This draft has been open since 2023 and `self_contained_coordinator.py` has diverged substantially since, so the branch no longer applies cleanly and has no CI. Closing for housekeeping — if the `process_args()` / `connect_redis_server()` extraction is still worth doing, a small fresh PR against current `main` would be very welcome.

Issue #244, opened 2023-07-03
[BUG]: tmp* folders are not deleting after execution of some tests


markovamaria:
**Describe the bug**
tmp* folders are not deleting after execution of some tests. 
For benchmark version 72,  home directory has 12 empty tmp* directories per run after execution.

**To Reproduce**
```
$ numactl --physcpubind=1 ./r_servers/redis-server --protected-mode no --port 6379 --dir run_server_logs --logfile run_server.log --save "" &
$ server_pid=$!
$ redis-benchmarks-spec-client-runner --db_server_host localhost --db_server_port 6379 --client_aggregated_results_folder ./run --flushall_on_every_test_start --flushall_on_every_test_end
$ kill -9 $server_pid 
```

**Expected behavior**
There is no empty tmp* directories in home directory after execution.

PR #245, opened 2023-08-29
Move temporary dir deletion to inner function, delete temp dir in skipped benchmarks

mpozniak95:
Probably connected with issue: https://github.com/redis/redis-benchmarks-specification/issues/244.

So far when some check resulted with skipping benchmark, it would keep an empty temporary directory in home directory. Deletion of temporary dir should happen in such situation before continuing to the next benchmark.

This PR moves temporary directory deletion to inner function in `process_self_contained_coordinator_stream` and invokes it before skipping any benchmark


mpozniak95:
I will open this PR from new branch located in this repository since looks like github jobs run for PR from forks doesn't have access to github secrets.

PR #216, opened 2023-02-22
Added conversion for str for git timestamps

markovamaria:
Added conversion to avoid error during testing triggering:
```
$ redis-benchmarks-spec-cli --use-branch --from-date 2023-02-11  --redis_port 12010 --redis_host *** --redis_pass *** --redis_user default
Traceback (most recent call last):
  File "/home/ubuntu/.local/bin/redis-benchmarks-spec-cli", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/.local/lib/python3.8/site-packages/redis_benchmarks_specification/__cli__/cli.py", line 53, in main
    trigger_tests_cli_command_logic(args, project_name, project_version)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/redis_benchmarks_specification/__cli__/cli.py", line 259, in trigger_tests_cli_command_logic
    ) = request_build_from_commit_info(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/redis_benchmarks_specification/__common__/builder_schema.py", line 151, in request_build_from_commit_info
    raise Exception(
Exception: Type of field commit_datetime is not bytes, string, int or float. Type (<class 'datetime.datetime'>). Value=2023-02-21 18:58:55+02:00
```

PR #248, opened 2023-09-04
Move temporary dir deletion to inner function, delete temp dir in skipped benchmarks

mpozniak95:
Opened again from branch created in this repository instead of forked branch.

Probably connected with issue: https://github.com/redis/redis-benchmarks-specification/issues/244.

So far when some check resulted with skipping benchmark, it would keep an empty temporary directory in home directory. Deletion of temporary dir should happen in such situation before continuing to the next benchmark.

This PR moves temporary directory deletion to inner function in process_self_contained_coordinator_stream and invokes it before skipping any benchmark

### DIFF
--- a/redis_benchmarks_specification/__self_contained_coordinator__/self_contained_coordinator.py
+++ b/redis_benchmarks_specification/__self_contained_coordinator__/self_contained_coordinator.py
@@ -912,9 +912,7 @@ def process_self_contained_coordinator_stream(
                                     logging.critical("Printing redis container log....")
                                     print("-" * 60)
                                     print(
-                                        redis_container.logs(
-                                            stdout=True, stderr=True, logs=True
-                                        )
+                                        redis_container.logs(stdout=True, stderr=True)
                                     )
                                     print("-" * 60)
                                 test_result = False
--- a/redis_benchmarks_specification/__runner__/runner.py
+++ b/redis_benchmarks_specification/__runner__/runner.py
@@ -359,6 +359,19 @@ def process_self_contained_coordinator_stream(
     resp_version=None,
     override_memtier_test_time=0,
 ):
+    def delete_temporary_files(
+        temporary_dir_client, full_result_path, benchmark_tool_global
+    ):
+        if preserve_temporary_client_dirs is True:
+            logging.info(f"Preserving temporary client dir {temporary_dir_client}")
+        else:
+            if "redis-benchmark" in benchmark_tool_global:
+                if full_result_path is not None:
+                    os.remove(full_result_path)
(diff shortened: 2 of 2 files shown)


redis_benchmarks_specification/__self_contained_coordinator__/self_contained_coordinator.py
```diff
@@ -912,9 +912,7 @@ def process_self_contained_coordinator_stream(
                                     logging.critical("Printing redis container log....")
                                     print("-" * 60)
                                     print(
-                                        redis_container.logs(
-                                            stdout=True, stderr=True, logs=True
```
mpozniak95, line 916: Also there was an issue with this line in contained_coordinator. When benchmark raised an exception for any reason we tried to print container logs but it resulted with: `TypeError: ContainerApiMixin.logs() got an unexpected keyword argument 'logs'`. According to https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.logs parameter `logs `for method `logs()` doesn't exist. This led to skipping 'teardown' stage which led to temporary folders being left with the redis-server artifact inside.

mpozniak95: With that change instead we get:
```
2023-09-13 01:33:49 CRITICAL Some unexpected exception was caught during local work. Failing test....
2023-09-13 01:33:49 CRITICAL <class 'KeyError'>
2023-09-13 01:33:49 INFO Tearing down setup
2023-09-13 01:33:49 INFO When trying to stop DB container with id a9b45d05a8ba306464d79d3d0deeaa2a72dd7d8761b838aefe5d318326e1cf5e and image <Image: 'debian:buster'> it was already stopped
2023-09-13 01:33:49 INFO Removing temporary dirs /root/tmp7rhog4yu and /root/tmpkcu7fzqu
```
And temporary dirs are removed

markovamaria (changes requested):
Hi Marcin, thank you for PR!

Could you please fix integration test failure? This is failing - test_run_local_command_logic_oss_cluster  according to log. 

You can use these PRs for test's run reference: 
- https://github.com/redis/redis-benchmarks-specification/pull/211
- https://github.com/redis/redis-benchmarks-specification/pull/216

Let me know in case of questions.

mpozniak95:
I think that that wasn't fault of my PR, there was an "Connection Refused" exception when trying to connect to Redis. I re-run test and it worked.

markovamaria (approved):

markovamaria:
@filipecosta90 , could you please check and merge if you are agree?

PR #288, opened 2025-05-08
Prevent stats to fail on unknown commands

paulorsousa:


fcostaoliveira (approved):
LGTM. Thank you @paulorsousa!

PR #319, opened 2025-10-15
Topologies

slice4e:
Creates a unique filename for each test result, when the test is configured to run with multiple topologies. Currently, all the result files show a hardcoded "oss-standalone" which is inaccurate. This pull request fixes this issue and result files show correct topology.


fcostaoliveira:
thank you @slice4e !

PR #343, opened 2026-01-19
Enable selecting regression/improvements/simplified comparison output

fcostaoliveira:


fcostaoliveira:
Closing as superseded. The substantive features here have since landed in `main` independently:

- `print_regressions_only` / `print_improvements_only` filtering is wired in `compare.py` (the conditional regression/improvement table sections).
- The `metric_mode` higher-better / lower-better sign logic is present in `compare.py`.
- `simplify_table` gating of the GROUP/COMMAND/full-results CSV blocks is present.

The branch is ~5 months old and now conflicts across all of `compare.py`'s changed regions, and the diff also carries debug `logging.info` blocks and hardcoded environment-specific filters that shouldn't merge.

The one remaining useful bit — converting the `--print-regressions-only` / `--print-improvements-only` flags from `type=bool` (an argparse footgun) to `action="store_true"` — is better re-cut as a small fresh PR against current `main`. Happy to open that separately if still wanted.

PR #374, opened 2026-04-15
stream 70-30 benchmark: use --test-time 120 for stable variance

paulorsousa:
Replace `-n 1000` with `--test-time 120` to ensure steady-state measurement.

With `-n 1000` (200K total requests at ~127K ops/s), the test finishes in ~1-2 seconds, making it extremely sensitive to startup jitter — observed **99.6% std.dev** across runs on ARM.

The `1k-consumers` variant of this same test already uses `--test-time 120` and achieves **0.1% variance**.

### Change
- `-n 1000` → `--test-time 120` (120-second steady-state run)
- Updated description to reflect the new measurement mode

---

> [!NOTE]
> **Low Risk**
> Changes are limited to benchmark YAML specifications (no runtime code changes), with primary impact being longer and more stressful benchmark runs that may affect CI/resource usage.
> 
> **Overview**
> Adds two new stream GC-focused memtier benchmark specs (`memtier_benchmark-stream-50-cgroups-xadd-xreadgroup-xtrim-{gc,aggressive-gc}.yml`) that fully populate 50 consumer-group PELs and include a 50/20/30 mix of `XADD`/`XREADGROUP`/`XTRIM` to force GC compaction under active consumption.
> 
> Replaces the previous `stream-50-cgroups` `xadd`+`xtrim`-only suites (which only preloaded 2k PEL entries) and updates `memtier_benchmark-stream-concurrent-xadd-xreadgroup-70-30.yml` to run for `--test-time 120` (instead of `-n 1000`) with refreshed description for steadier measurements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5996773712ee1682d9dbc8d4a1b78a7e0abb8338. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-stream-50-cgroups-xadd-xreadgroup-xtrim-aggressive-gc.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-stream-50-cgroups-xadd-xreadgroup-xtrim-aggressive-gc.yml
@@ -1,12 +1,12 @@
 version: 0.4
-name: memtier_benchmark-stream-50-cgroups-xadd-xtrim-gc
+name: memtier_benchmark-stream-50-cgroups-xadd-xreadgroup-xtrim-aggressive-gc
 description: >
-  Tests stream GC/compaction overhead with 50 consumer groups.
-  Pre-loads 100K stream entries and creates 50 consumer groups, each reading all entries into their PEL.
-  The benchmark runs 50% XADD, 20% XREADGROUP (advancing min_cgroup_last_id), and 30% XTRIM MAXLEN ~ 50000.
-  XREADGROUP is critical — without it min_cgroup_last_id never advances and streamEntryIsReferenced()
-  short-circuits before reaching the PEL scan. With active consumption, trimmed entries fall through
-  to the O(C) PEL scan across all 50 consumer groups.
+  Tests stream GC cleanup under aggressive trimming with 50 consumer groups and populated PELs.
+  Pre-loads 100K stream entries, creates 50 consumer groups, and reads all entries into
+  each group's PEL. The benchmark runs 50% XADD, 20% XREADGROUP (advancing min_cgroup_last_id),
+  and 30% XTRIM MAXLEN ~ 1000 to force frequent GC compaction.
+  XREADGROUP advances min_cgroup_last_id so trimmed entries fall through to the O(C) PEL scan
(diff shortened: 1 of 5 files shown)


fcostaoliveira:
Closing this one as stale/superseded — but the GC-benchmark idea is worth re-cutting against current `unstable`, so here's a concrete path rather than a flat no.

**Why close now**

1. **Headline change already landed.** The `-n 1000 → --test-time` switch on `stream-concurrent-xadd-xreadgroup-70-30` is already on `unstable` (the 70-30 spec uses `--test-time`), so that part is moot, and the branch now conflicts with the independently-reworked stream specs (the two files this PR deletes still exist on `unstable` and were reworked since).

2. **The GC specs don't currently exercise the path they describe.** On current `unstable` (`t_stream.c`):
   - `XTRIM stream-key MAXLEN ~ 1000` with no strategy keyword defaults to **`KEEPREF`**, whose trim path deletes nodes directly and never calls `streamEntryIsReferenced()` / `streamCleanupEntryCGroupRefs()`. Those are only reached under `ACKED` / `DELREF`. To hit the GC path the client needs `XTRIM ... ACKED` (or `XACKDEL` / `XDEL ... DELREF`).
   - `streamEntryIsReferenced()` now resolves references via the `cgroups_ref` rax index (`raxFind`, ~O(log N)), not the O(C) all-groups PEL scan the descriptions assume.

3. **The preload is trimmed away.** The 50 `XGROUP CREATE` + `XREADGROUP COUNT 100000` init builds 100K-entry PELs, but `XTRIM MAXLEN ~ 1000` collapses the stream to ~1000 within seconds and the client `XREADGROUP ... >` reads only *new* entries — so the populated PELs are aged out and steady state no longer matches the "50 fully-populated PELs under GC" claim.

(The red CI is unrelated: it's a stale `black --check` formatting drift at that commit, `commands[0]` in tox — `Validate SPEC fields` passes and the YAMLs parse fine.)

**If you want to re-cut it**

A 50-consumer-group GC benchmark is genuinely useful — rebased on current `unstable`, with `XTRIM ... ACKED` (or `XACKDEL`/`DELREF`) so the GC functions are actually called, a non-draining steady state that keeps the PELs populated, a metric-of-interest in each `description`, and a title that matches the change. Happy to help shape that.

PR #434, opened 2026-07-02
test-suites: pipelined score-collision ZINCRBY throughput spec (1M elements, pipeline-10)

fcostaoliveira:
## What

Adds a **throughput** counterpart to the pipeline-1 score-collision ZINCRBY spec added in #430:

`memtier_benchmark-1key-zincrby-1M-elements-score-collision-pipeline-10`

Bumps the package version to `0.3.26`.

## Why

The existing `…-score-collision-pipeline-1` spec is a per-op **latency probe** (`-c 1 -t 1 --pipeline 1`). It successfully drives every ZINCRBY onto the skiplist's *equal-score-run* path — all 1M members start at score `0`, so the live scores stay concentrated in a narrow translating integer band and each reinsert lands into a dense equal-score run, where `zslCompareWithNode` falls back to member-string `sdscmp` instead of the cheap early-terminating score compare.

But at `--pipeline 1` the single-connection Ops/sec is bounded by the **per-op kernel `write()` syscall**, which masks the per-op skiplist-walk CPU cost. So the pipeline-1 spec cannot show whether a change to the equal-score-run walk moves throughput.

This variant saturates the single-threaded server core (`-t 4 -c 50` = 200 connections, `--pipeline 10`, `--test-time 180`), amortizing the syscall ~10× so the equal-score-run walk becomes the binding constraint and surfaces as an **Ops/sec** delta.

## Design notes

- **Reuses the #430 dataset** (`dataset_name: 1key-zset-1M-elements-score-collision`) — the preload is byte-identical, so the 1M-member score-0 artifact is shared.
- **`~99.8%` equal-score-run firing** for the whole 180s window: modelling per-member touch counts as Poisson, the score band stays tens-of-thousands deep even at the end of the run; only a ~0.2–1% leading-edge tail (members pushed to a brand-new maximum score with no neighbor yet) takes the ordinary path. The description states this honestly rather than claiming 100%.
- **Keyspace aligned**: client `--key-maximum 999999` matches the preload's exact member range (IDs `0..999999`), so the `R`-uniform selection only ever targets preloaded members.
- **Encoding**: skiplist+hashtable (1M elements > `zset-max-listpack-entries` 128); members are numeric-string SDS compared via `sdscmp`.
- Metric of interest stated as **Ops/sec** (saturated throughput), matching the load model.

## Why `--pipeline 10` (and why one spec, not a matrix)

`10` is not arbitrary — it's the repo's established depth for single-key sorted-set **throughput** probes (`…-zscore-pipeline-10`, `…-zcard-pipeline-10`, `…-zremrangebyscore-pipeline-10` all use `--pipeline 10`). Together with the existing `--pipeline 1` score-collision sibling from #430, this forms the natural two-point latency→throughput pair for the same dataset, so no new matrix is needed to justify the value.

The spec is a **measurement instrument added ahead of the change it measures** — the same ordering used for the pipeline-1 collision spec in #430. It is self-contained and does not depend on any unmerged code: it simply gives the equal-score-run ZINCRBY path a saturating throughput probe that the syscall-bound pipeline-1 spec cannot provide. Even absent any future optimization, it is valid coverage of a dense-equal-score-run ZINCRBY workload.

## Validation

Key-range semantics confirmed empirically against the `redislabs/memtier_benchmark:edge` image used by the runners: `R` selection over `--key-maximum N` covers `[0, N]` **inclusive**, and the sequential preload (`-n 1000000`, `--key-maximum 1000000`) creates members `0..999999`. Client `--key-maximum 999999` therefore selects **exactly** the 1M preloaded members — no ZINCRBY ever lands on a non-preloaded id (which would create a member outside the score band).

## Test plan

- `tested-commands: [zincrby]`, `tested-groups: [sorted-set]`, `keyspacelen: 1`.
- oss-standalone; amd64 + arm64 + dockerhub build variants.

---

> [!NOTE]
> **Low Risk**
> Adds only a benchmark YAML definition with no runtime or Redis server code changes.
> 
> **Overview**
> Adds **`memtier_benchmark-1key-zincrby-1M-elements-score-collision-pipeline-10`**, a throughput companion to the existing `…-pipeline-1` score-collision **ZINCRBY** spec on the shared **`1key-zset-1M-elements-score-collision`** dataset (1M members, all score 0).
> 
> The workload still issues random **`ZINCRBY lb 1`** against that single key so updates stay on the skiplist **equal-score-run** path (`sdscmp` on members). The client load model changes to **`-t 4 -c 50 --pipeline 10`** over **180s** so syscall overhead is amortized and **Ops/sec** reflects skiplist-walk CPU rather than per-op `write()` limits. Preload, topology (**oss-standalone**), and build variants match the existing collision pattern; **priority 67**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6f02b98b1fe11bc7608ffe72e13e56f2535acaf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Ran this through a 7-lens adversarial review. Findings:

**Fixed:**
- Stale version bump: this branch bumped `pyproject.toml` to 0.3.26, main is now at 0.3.34 (several releases since). Resolved via merge (not rebase, to avoid disturbing the open PR's history) — was showing `mergeable: CONFLICTING` before this, now clean.
- The description's `~99.8%` / `~0.2-1%` fast-path-firing split was asserted with no derivation, and doesn't obviously square with the load model this same spec claims (saturating throughput implies a touch-count mean well above the "single-digit" the percentages seem to assume). Softened to qualitative language — the mechanism description itself (equal-score members take the `sdscmp` skiplist path) holds up under review, just not those specific digits.

**The bigger finding — this fills the missing cross-reference a review lens flagged:** this spec isn't speculative infrastructure for a hypothetical future change. It's already served its purpose — [`approaches/zset-update-score-equal-neighbors.md`](https://github.com/redis-performance/redis-agent-workspace/blob/main/approaches/zset-update-score-equal-neighbors.md) (internal) shows this exact spec was used on 2026-07-04 to test a `zslUpdateScore` equal-score-neighbor fast path and **reject it** (+1.32% x86, −3.12% ARM — sub-threshold, ARM regression). The value of merging this now isn't "measure a pending change" — it's permanent regression coverage of a hot path someone will likely revisit (Redis still uses skiplist; nothing changed on the Redis side).

**Confirmed correct, not changed:**
- Preload/client key-range pairing (`P` pattern's exclusive upper bound vs `R`'s inclusive) is exactly right — 7 empirical checks (DBSIZE, ZCARD, ZSCORE samples, live drift check) found no off-by-one. Notably, its `pipeline-1` sibling already on `main` *does* have a 1-in-a-million off-by-one this file avoids.
- Encoding claim ("skiplist+hashtable" at 1M elements) verified against `OBJECT ENCODING` and `zset-max-listpack-entries` (128, unlike this repo's surprising 512 default for hashes) — accurate.
- `keyspacelen`, `tested-commands`, topology, build-variants, resource sizing, `:edge` flag compatibility — all checked, all clean.

PR #436, opened 2026-07-03
test-suites: add ZRANDMEMBER count coverage on a 10K-element skiplist zset (CASE 3)

fcostaoliveira:
## Summary

Adds a single-key benchmark exercising `ZRANDMEMBER key <count>` on a **skiplist-encoded** sorted set, complementing the existing SRANDMEMBER count coverage (#435).

- **Preload:** one key `zset:10K:elements` with 10,000 distinct string members (skiplist, well past `zset-max-listpack-entries`).
- **Client:** `ZRANDMEMBER zset:10K:elements 5000`, `--pipeline 4`, 120s.

## What it exercises

`5000 * ZRANDMEMBER_SUB_STRATEGY_MUL (3) > 10000`, so `zrandmemberWithCountCommand` takes its **CASE 3** strategy: build an auxiliary dictionary holding the whole zset, then remove random elements down to the requested count. Only skiplist-encoded zsets reach CASE 3 (listpack is served earlier), and in the generic path every inserted member is a heap-allocated SDS copy — so this path is dominated by per-element SDS allocation + free churn. This spec gives that allocation-heavy large-fraction path stable `Ops/sec` coverage, alongside the sibling SRANDMEMBER count specs.

## Notes

- Read-only client phase (encoding + size stable across the measurement window).
- `tested-commands: [zrandmember]`, `tested-groups: [sorted-set]`; amd64 + arm64 build variants; encoding documented in the description.

---

> [!NOTE]
> **Low Risk**
> Benchmark YAML only; no application or Redis server code changes.
> 
> **Overview**
> Adds a new memtier benchmark spec that measures **`ZRANDMEMBER`** with **count 5000** on a single **10K-member skiplist** zset, mirroring the existing SRANDMEMBER large-count CASE 3 coverage for sorted sets.
> 
> The workload preloads `zset:10K:elements` via `ZADD`, then runs read-only **`ZRANDMEMBER zset:10K:elements 5000`** with pipeline 4 for 120s. The description documents that this count triggers **`zrandmemberWithCountCommand` CASE 3** (full auxiliary dict + per-member SDS alloc/free on skiplist), tagged under **`sorted-set`** / **`zrandmember`**, with amd64/arm64 and dockerhub build variants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc83d8711ab6829305852d7746e21d0c46ecbda5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Ran this through a 7-lens adversarial review. The technical claims in the description held up unusually well — verified line-by-line against `redis/src/t_zset.c` (`ZRANDMEMBER_SUB_STRATEGY_MUL` value, the CASE 2.5/CASE 3 control flow, `zuiNewSdsFromValue` as the exact allocation site) and everything checked out, including the subtler claim that listpack zsets are intercepted by CASE 2.5 *before* the CASE 3 size/count ratio is ever evaluated.

**Two things fixed:**
- The `--pipeline 4` description claimed it was chosen to bound in-flight reply memory — that framing ignored memtier's default 200 connections (`-t 4 -c 50`), so the real bound is dominated by connection count, not pipeline depth. Measured client CPU directly instead of re-deriving the claim: cpuset-limited to 4 cores, the client sits at ~2% utilization for the whole run. Replaced the memory-bounding claim with that measurement.
- Checked whether this suite needed the same client-thread fix as three sibling suites earlier this session (#485/#486/#487, all measured 81-91% client-CPU-bound) — it doesn't. This workload hits a single fixed key with no per-call variation, so the server's CASE 3 dict-build/teardown dominates and the client stays idle. Restraint here is backed by the same measurement discipline as the suites that *did* need fixing.

**Why this one matters more than a routine coverage PR:** this spec isn't speculative — it's the benchmark that already validated a real, accepted optimization. [`approaches/zrandmember-count-borrow-sds.md`](https://github.com/redis-performance/redis-agent-workspace/blob/main/approaches/zrandmember-count-borrow-sds.md) (internal) shows it fleet-confirmed +10.8% x86 / +11.4% arm64 for redis/redis#15421 (adversarial gate passed, currently in review upstream), and that approach's own "Next" section names *this PR* as the last step before permanent fleet registration. Merging this is what lets that win stay guarded against regression going forward.

Issue #433, opened 2026-07-01
Add benchmark coverage for RESTORE command


fcostaoliveira:
The current CE Performance suite has no dedicated coverage for the `RESTORE` command path (nor for `DUMP`). RESTORE is a performance-sensitive data-plane command — it is on the hot path for cluster resharding, key migration flows (`DUMP` + `RESTORE`), and any tooling that moves keys between shards or clusters.

Without a targeted benchmark we cannot:
- detect throughput/latency regressions on the RESTORE insertion path,
- validate improvements to the new-key vs. replace-key branches of `restoreCommand`,
- track dbAdd/dbDelete cost changes on this path across releases.

Suggested initial spec (memtier / oss-standalone, `generic` group):
- **Warm-up:** pre-populate N string/hash keys (small values), then `DUMP` each to build a payload set.
- **Steady state:** pipeline `RESTORE <key> 0 <payload>` against fresh key names — the common "load new keys" case. Variants worth having:
  - `RESTORE ... REPLACE` against existing keys (the replace path).
  - `RESTORE` with a mix of small (10B) and medium (1KB) payloads for allocator sensitivity.
- Report the standard ops/sec + p50/p99 alongside the existing generic-group specs.

Happy to iterate on payload sizes / value-type mix if there is a preferred shape.

mpozniak95:
I think this can be closed now :)

PR #439, opened 2026-07-10
Add RESTORE benchmark coverage (issue #433)

mpozniak95:
Add 16 memtier test-suites for the RESTORE command (generic group, oss-standalone), covering the string and hash value types across small and 1KB payloads, the load-new-keys (dbAdd) and REPLACE (dbDelete+dbAdd) insertion paths, and pipeline depths 1 (latency) and 10 (throughput).

Payloads are real DUMP blobs generated via SET/HSET + DUMP on Redis 7.4 (RDB version 12), embedded as hex-escaped literals in the memtier --command. hash-small exercises the listpack decode path; hash-1KB the hashtable decode path.

---

> [!NOTE]
> **Low Risk**
> Benchmark specification YAML only; no runtime or server code changes. CI may use more memory/time on new load-new-keys throughput cases.
> 
> **Overview**
> Adds **16 new memtier benchmark specs** for the `restore` command (`generic` / `oss-standalone`), addressing issue #433.
> 
> Coverage spans **string** (small EMBSTR, 1KB RAW) and **hash** (small listpack, ~1KB hashtable) DUMP payloads from Redis 7.4 / RDB v12, embedded as hex in `--command`. Each type has **load-new-keys** (`dbAdd` on fresh keys, partitioned key range with no-wrap comments) and **replace-existing** (`RESTORE ... REPLACE` over a 1000-key preload with `keyspacelen` check), at **pipeline 1** (latency) and **pipeline 10** (throughput). Load specs scale DB memory requests for growing keyspaces; throughput variants use higher client concurrency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e46bebc27fccc4c2c8339858a7ef2d453c24f02e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mpozniak95:
I couldn't find an easy way to DUMP a key in preload and pass that dump into the benchmark phase — I think that the framework doesn't carry values between the two (init_lua's output isn't captured anywhere). So the payloads here are fixed literals, generated once by running SET/HSET + DUMP on Redis 7.4 and embedded hex-escaped in the --command. On the wire it's identical to a live DUMP, so RESTORE still hits the real deserialization path.

Is it fine, or would you rather I extend the benchmark to pass variables from preload into the main phase?

mpozniak95:
@fcostaoliveira or @paulorsousa please review when you have time :)

fcostaoliveira (commented):
Reviewed against the memtier design checklist, with the payload mechanics verified empirically on the exact fleet client image (`redislabs/memtier_benchmark:edge`) against a scratch server — this is the part that can't be judged by reading:

**Verified end-to-end (all pass):**
- The `\x..` hex escapes decode correctly through edge memtier's `--command` parser — RESTORE succeeds (no checksum errors), keys are created with the right content.
- **string-small** → `embstr`, value `AAAAAAAAAA`; **string-1KB** → `raw`, `STRLEN` 1024 (the incompressible-payload choice correctly avoids embedding an LZF-compressed blob, which would have silently added `lzf_decompress` to the measured path); **hash-small** → `listpack`, 3 fields; **hash-1KB** → `hashtable`, 8 fields × 120B values — every encoding claim in the descriptions matches reality.
- **REPLACE path**: runs clean against the preloaded keyspace, DBSIZE stays constant — no BUSYKEY trap.
- **Preload off-by-one**: `--key-maximum 101 -n allkeys` created exactly 100 keys (exclusive upper bound on the `--command` path), so `keyspacelen: 1000` with preload max 1001 + client R over [1,1000] is exactly aligned.

**Suggestions (non-blocking):**
1. **Memory honesty on the pipeline-10 load-new-keys variants.** `-c 50 -t 4` at pipeline 10 for 180s will create tens of millions of keys; the hash-1KB variant plausibly grows the dataset by tens of GB during the measured window, while `dbconfig.resources.requests.memory` says `1g`. It's advisory on our runners, but I'd bump the requests to honest values (e.g. 8g for the string p10 cells, 64g for hash-1KB p10) and add one description sentence noting the run measures `dbAdd` against a continuously growing dict (rehash cost drifts across the window — inherent to load-shape specs, worth stating).
2. **No-wrap margin comment.** The 1e9 key range across 200 P-partitioned connections gives ~5M keys/connection vs ≲1M ops/connection in 180s — safely no wrap. Worth a short YAML comment, because if someone later raises `-c/-t` or `--test-time` enough to wrap a slice, the spec silently flips to the BUSYKEY error path (there's no REPLACE on these cells).
3. **replace-existing working set** (1000 keys × small payloads) is fully cache-resident — good for isolating the replace code path from memory effects, but worth the word "cache-resident" in the description so nobody reads it as a memory-pressure result.
4. Nit: `cpus: '4'` requested for `-t 1` single-connection latency probes.

CI is green (incl. the spec-fields validator). With #1 addressed this looks ready.

fcostaoliveira:
Applied the review suggestions directly in e46bebc (maintainer edit): honest per-cell memory requests on the load-new-keys variants (sized to actual keyspace growth per pipeline depth) plus the growth-drift note, the no-wrap invariant documented as a YAML comment next to each load clientconfig, the cache-resident note on the replace-existing cells, and the pipeline-1 client cpus nit. All 16 files re-validated (parse + spec-fields).

fcostaoliveira (approved):
Payload mechanics verified empirically on the fleet client image (all four DUMP blobs decode and land in the claimed encodings; REPLACE path clean; preload off-by-one aligned) and all review suggestions applied in e46bebc. CI 8/8 green.

Issue #422, opened 2026-06-14
Add benchmark coverage for LSET (list element replacement path)


fcostaoliveira:
The `list` group covers RPUSH/LPUSH/LPOP/RPOP/LRANGE/LLEN/LINDEX/LINSERT/LREM/LPOS/LMOVE/RPOPLPUSH today, but there is no benchmark exercising **LSET**. Grepping the spec directory:

```
$ grep -rliE '\blset\b' redis_benchmarks_specification/test-suites/
(no results)
```

LSET is the canonical "in-place element replacement" command on a list, and its implementation in `t_list.c:lsetCommand` runs an encoding-conversion call on every invocation (listpack ↔ quicklist), so its per-op cost is non-trivial and a regression in that path is currently invisible to the suite.

Suggested coverage, in increasing scope:

1. **Mid-sized listpack, LSET in place** — replace at index 0 with a same-sized small value. Covers the "no encoding change" hot path on listpack.
2. **Listpack growing into quicklist on LSET** — replace a small element with a value above `list-max-listpack-size` (default −2, so ~8 KiB), forcing a one-shot conversion. Covers the GROWING branch.
3. **Quicklist LSET with potential shrink** — pre-load a quicklist (large value or many elements), then LSET-in a small value to exercise the SHRINKING branch in `listTypeTryConversion`.

A natural shape for (1) would mirror the existing `memtier_benchmark-1key-list-100-elements-lrange-all-elements-pipeline-10`-style spec but use `--command "LSET __key__ __key__ __data__"` instead of LRANGE. Suite-wide, three specs at typical priorities should be enough to land LSET in the regular CE-Performance run.

Happy to follow up with the YAML if helpful.

PR #441, opened 2026-07-14
test-suites: add benchmark coverage for LSET command

mpozniak95:
Adds three test-case specs for LSET (issue #422), which was the only list command without benchmark coverage. Its per-call encoding-conversion check (t_list.c:lsetCommand) meant regressions were invisible to the suite.

Scenarios:
- 100-element listpack, LSET in place (no encoding change hot path)
- 10-element listpack LSET with 9000B value forcing listpack->quicklist (GROWING branch)
- 10K-element quicklist LSET with small value (SHRINKING branch)

Verified locally: all three run end-to-end against Redis 7.4.9, encodings confirmed via OBJECT ENCODING, LSET traffic confirmed in result stats.

---

> [!NOTE]
> **Low Risk**
> Only new benchmark YAML definitions; no production code, auth, or infrastructure changes.
> 
> **Overview**
> Adds **three new memtier_benchmark test-suite YAML specs** so **`LSET`** is covered in the Redis benchmarks suite (closes the gap called out in #422).
> 
> The specs target **`listTypeTryConversion`** behavior on **`LSET`**: a **100-element listpack** with in-place small replacement (encoding unchanged), a **10-element listpack** forced to **quicklist** via a **9000B** value (GROWING conversion), and a **~10K-element quicklist** with a small **LSET** (SHRINKING check path). The large-list case reuses the existing **`1key-list-10K-elements-string`** preload pattern used by sibling list benchmarks.
> 
> All three follow the usual **oss-standalone** topology, **gcc/dockerhub** build variants, and **priority 34**—benchmark-only additions with no runner or Redis code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 518da2b89e6e2098b011b097299ac99236016eec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-growlist-10-elements-lset-listpack-to-quicklist.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-growlist-10-elements-lset-listpack-to-quicklist.yml
@@ -1,5 +1,5 @@
 version: 0.4
-name: memtier_benchmark-1key-list-10-elements-lset-listpack-to-quicklist
+name: memtier_benchmark-1key-growlist-10-elements-lset-listpack-to-quicklist
 description: 'Runs memtier_benchmark, for a keyspace length of 1 LIST key. The LIST starts as a small
   listpack (10 elements) and we LSET an element with a value of 9000 Bytes, which is above the
   list-max-listpack-size -2 (8 KiB) node limit and therefore forces a listpack -> quicklist
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10-elements-lset-listpack-to-quicklist.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10-elements-lset-listpack-to-quicklist.yml
@@ -1,0 +1,41 @@
+version: 0.4
+name: memtier_benchmark-1key-list-10-elements-lset-listpack-to-quicklist
+description: 'Runs memtier_benchmark, for a keyspace length of 1 LIST key. The LIST starts as a small
+  listpack (10 elements) and we LSET an element with a value of 9000 Bytes, which is above the
+  list-max-listpack-size -2 (8 KiB) node limit and therefore forces a listpack -> quicklist
+  conversion. This exercises the GROWING branch of the per-call encoding-conversion check
+  (listTypeTryConversionRaw in t_list.c). The first LSET performs the actual conversion; every
+  subsequent LSET still pays the encoding-conversion check on a quicklist. Encoding: listpack -> quicklist.'
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  check:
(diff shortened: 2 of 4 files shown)


fcostaoliveira (changes requested):
Nice set — the listpack/quicklist boundary is chosen correctly (verified `9000B` genuinely crosses the `8192B` default threshold, and the 100-element in-place case correctly stays listpack). Two small things before merge:

**`memtier_benchmark-1key-list-10K-elements-lset-quicklist.yml` preloads 9999 elements, not 10,000.** The file name, description, and `dataset_description` all say 10K/10000 — measured `LLEN` after running the preload args as written comes out to 9999 (`--key-maximum 10000` with the preload's pattern is exclusive of the upper bound here). Doesn't change the encoding claim (both numbers are far past the quicklist threshold either way), but worth fixing the `--key-maximum` to actually create 10,000 elements, or correcting the three "10K"/"10000" references to 9999 — whichever's less churn.

**Minor cleanup:** the growth-case spec (`listpack-to-quicklist`) sets `--command-key-pattern R` but the command targets a literal key (`growlist:10`) with no `__key__` token, so the flag is dead/unused. Harmless, but worth dropping for clarity.

mpozniak95:
I changed in the description that it is really a 9999 instead of 10k

fcostaoliveira (dismissed):
Confirmed — the updated description accurately reflects 9999 elements with a clear explanation of why (the `P`-pattern's `-n allkeys` behavior), and the "10K" naming is reasonably justified as the shared family/dataset label matching the LINDEX/LPOS/LINSERT-LREM siblings. That resolves the finding. Thanks for the quick turnaround.

fcostaoliveira:
Reviewed this PR (7 adversarial passes + empirical verification). Two things addressed, both pushed to this branch (maintainer-edits enabled):

1. **Bugbot's dataset_name collision finding** — real. The LSET listpack→quicklist spec reused `dataset_name: 1key-list-10-elements` from the LRANGE siblings while seeding different data (different key, different values, different seed command). Renamed to `1key-growlist-10-elements` (and the file/`name:` field to match) so dataset-consistency tooling doesn't conflate them.

2. **A new finding**: the 10K-elements LSET spec's `dataset_description` says the shared preload actually yields 9,999 elements, not 10,000 — which contradicts what the 7 already-merged sibling specs (lindex/lpos/linsert-lrem) claim for the *same* preload command. Verified empirically against a real `redis-server` + `memtier_benchmark:edge`: `LLEN` really is 9,999. So this PR's own claim is correct, and it's the pre-existing merged specs that are wrong — filed as #509 to track fixing those separately, since it's out of scope for this PR and affects already-shipped specs. Nothing to change here.

Everything else checked out (encoding-conversion math against `t_list.c`/`quicklist.c`, key ranges, empirical validation). Looks good to merge once CI is green.

Issue #400, opened 2026-05-31
Add benchmark coverage for CLIENT LIST and CLIENT KILL (filter paths)


fcostaoliveira:
The existing `connection` test group covers `CLIENT INFO`, `HELLO`, and `PING` (see `memtier_benchmark-1Mkeys-string-client-info.yml`, `memtier_benchmark-connection-hello*.yml`, `memtier_benchmark-nokeys-connection-ping-pipeline-10.yml`). There is no coverage for `CLIENT LIST` or `CLIENT KILL`, both of which exercise distinct paths from `CLIENT INFO`:

- `CLIENT LIST` iterates the full client list and serializes per-client info via `getAllClientsInfoString` / `catClientInfoString` — cost scales with the number of connected clients and the number of fields rendered, neither of which `CLIENT INFO` (current connection only) exercises.
- `CLIENT KILL` walks the client list and matches against a filter set (`ID`, `ADDR`, `LADDR`, `USER`, `TYPE`, `SKIPME`, …) — the filter matching loop is a separate code path.

These commands are commonly used by tooling (RedisInsight, proxies, sentinel-style monitors) and changes to the filter / serialization paths (e.g. additional filter keywords, optional fields) can cause measurable per-call cost shifts that the current `connection` group will not catch.

Suggested additions:
- A `CLIENT LIST` test with a sustained client population (e.g. 500–2000 idle connections pre-opened by `memtier_benchmark -c 50 -t N` clients) and a small `CLIENT LIST` polling tool measuring throughput / latency.
- A `CLIENT LIST TYPE normal` and `CLIENT LIST ID <list>` variant so the filter-match path is exercised.
- A `CLIENT KILL ADDR <addr>` no-match variant (kills nothing but walks the filter logic) to isolate the filter cost.

Both should sit in the `connection` test group and reuse the existing `string-client-info` dataset pattern.

PR #445, opened 2026-07-28
test-suites: add connection-group coverage for CLIENT LIST and CLIENT KILL filter paths

mpozniak95:
Adds four test-suites to the connection group covering code paths not exercised by the existing CLIENT INFO/HELLO/PING specs:

- CLIENT LIST: full client-list iteration and per-client serialization (getAllClientsInfoString / catClientInfoString), cost scales with the number of connected clients.
- CLIENT LIST TYPE normal and CLIENT LIST ID <ids>: the filter-match path.
- CLIENT KILL ADDR <addr> (no-match): walks the filter-matching loop without killing anything, isolating filter cost.

Each spec keeps a sustained population of ~1000 connections open (-c 250 -t 4) so the list/filter loops walk a large client set, and reuses the 1Mkeys string dataset pattern from string-client-info.

Refs redis/redis-benchmarks-specification#400

---

> [!NOTE]
> **Low Risk**
> Only new benchmark YAML definitions; no runtime or product code changes.
> 
> **Overview**
> Adds **four** memtier benchmark specs under the **connection** group to cover `CLIENT` list/kill paths that existing `CLIENT INFO` / `HELLO` / `PING` suites do not hit.
> 
> **`CLIENT LIST`** measures full-list iteration and per-client serialization with ~1000 sustained connections (`-c 250 -t 4`) on the shared 1M-key string preload. **`CLIENT LIST TYPE normal`** targets the TYPE filter walk (expected to track plain `CLIENT LIST` on oss-standalone for regression on filter/type code). **`CLIENT LIST ID`** uses random ids via `__key__` to exercise radix-tree lookup instead of list walks. **`CLIENT KILL ADDR`** with a documentation-only address isolates filter-matching cost with no kills.
> 
> All four reuse the `memtier_benchmark-1Mkeys-string-client-info` db/preload pattern, `oss-standalone`, and the same gcc/dockerhub build matrix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0f0c93b979e29340d20f9d200a73f7c088fd5ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira (changes requested):
Thanks for this — the CLIENT LIST/KILL coverage gap is real and the connection-count sizing/description quality here is solid. One finding from review before merge:

**`client-list-id`'s description doesn't match the code path it exercises.** `CLIENT LIST ID <id>` resolves via `lookupClientByID()` → `raxFind(server.clients_index, ...)` in `networking.c` — an O(log n) radix-tree point lookup, not a walk of the connection list. `getAllClientsInfoString()` (the actual list-walk function this spec's siblings exercise) is only invoked when that lookup misses. So this spec isn't measuring a cheaper corner of the same walk `CLIENT LIST`/`CLIENT LIST TYPE` exercise — it's a different subsystem (dict/rax lookup) entirely. Worth correcting the description to say what it actually measures (a point lookup by ID, hit or miss) rather than framing it as a walk-cost variant.

Separately, not a blocker: `CLIENT LIST TYPE normal` will likely read statistically identical to plain `CLIENT LIST` here, since every sustained connection in this benchmark is type `normal` — the filter never excludes anything. Not broken, just low incremental signal versus the stated "filter-match path" objective; up to you whether that's worth a note in the description or a follow-up.

The unrelated `test_spin_docker_cluster_redis` CI failure is a known cluster-formation timing flake, not caused by this diff.

Issue #456, opened 2026-07-31
use_git_timestamp is written as str(bool) and parsed with bool(str), so 'False' is always True — datapoints are unconditionally commit-dated


fcostaoliveira:
## Summary

`use_git_timestamp` is written to the stream as `str(bool)` and read back with `bool(str)`. Since every non-empty string is truthy, `"False"` parses as `True` — so the flag can never be `False` on the stream path, and result datapoints are unconditionally stamped with the commit timestamp rather than wall-clock time.

## Evidence

Producers write the stringified bool:

```python
# __common__/builder_schema.py:182
commit_dict["use_git_timestamp"] = str(use_git_timestamp)

# __builder__/builder.py:962
"use_git_timestamp": str(use_git_timestamp),
```

Consumers parse it with `bool()`:

```python
# __self_contained_coordinator__/build_info.py:20-21
if b"use_git_timestamp" in testDetails:
    use_git_timestamp = bool(testDetails[b"use_git_timestamp"].decode())

# __builder__/builder.py:413-414   (note: no .decode() here either)
if b"use_git_timestamp" in testDetails:
    use_git_timestamp = bool(testDetails[b"use_git_timestamp"])
```

Both are unconditionally truthy for the values actually on the wire:

```
>>> bool("False")     # build_info.py path
True
>>> bool(b"False")    # builder.py path — bytes, not even decoded
True
```

The builder then re-emits the now-`True` value at `builder.py:962`, so the corruption propagates from the commits stream to the builds stream.

## Impact

`use_git_timestamp` selects the timestamp used for exported datapoints:

```python
# __self_contained_coordinator__/self_contained_coordinator.py:~2551-2558
datapoint_time_ms = start_time_ms
if (use_git_timestamp is True and git_timestamp_ms is not None):
    datapoint_time_ms = git_timestamp_ms
```

So for any stream-driven run, datapoints land at the **commit's authored time**, never at ingestion time — regardless of what the trigger requested. Two consequences:

1. **`--use-git-timestamp false` is inert.** The flag exists (`__cli__/args.py`) and is accepted, but cannot take effect through the stream path.
2. **Time-range queries need care.** A datapoint written today for a commit from three weeks ago is dated three weeks ago. Any comparison whose `from`/`to` window is anchored on "recent" silently misses it. This is especially visible when benchmarking a release tag, where the tag's commit date can be well before the run.

[...]

Issue #455, opened 2026-07-31
--dockerhub path passes git_version into the git_hash positional; git_hash gets a version string and git_version is None


fcostaoliveira:
## Summary

On the `--dockerhub` trigger path, `generate_benchmark_stream_request()` is called with ~30 positional arguments, and the version string lands in the `git_hash` parameter. `git_hash` receives a version like `8.10.0` instead of a commit SHA, and `git_version` receives `None`.

## Evidence

Signature (`__builder__/builder.py:927`), positions 14-17:

```
  14. git_branch
  15. git_hash
  16. git_timestamp_ms
  17. git_version
```

Call site (`__cli__/cli.py:228`), counting positionals:

```python
build_stream_fields, result = generate_benchmark_stream_request(
    args.id,          #  1 id
    conn,             #  2 conn
    args.run_image,   #  3 run_image
    args.arch,        #  4 build_arch
    testDetails,      #  5 testDetails
    "n/a",            #  6 build_os
    [],               #  7 build_artifacts
    None,             #  8 build_command
    None,             #  9 build_config_metadata
    None,             # 10 build_image
    None,             # 11 build_vars_str
    None,             # 12 compiler
    None,             # 13 cpp_compiler
    None,             # 14 git_branch
    git_version,      # 15 -> git_hash          <-- version string into the hash parameter
    None,             # 16 git_timestamp_ms
    None,             # 17 -> git_version       <-- left None
    ...
```

The trailing comment on that line reads `# Pass extracted version`, so the intent was clearly `git_version`; the value simply lands one parameter short of it.

## Impact

Results are keyed by `git_hash` in the datasink — the `by.hash` series path is built from it. A version string in that field means dockerhub-triggered runs write `by.hash` series namespaced by something that is not a hash, while `by.version` gets nothing because `git_version` is `None`. So those datapoints are hard to correlate with any commit, and a `--baseline-hash` / `--comparison-hash` comparison cannot find them.

I have not reproduced this end-to-end against the datasink — the finding is from the signature/call-site mismatch, which is unambiguous. Worth confirming how much dockerhub-path data exists before deciding urgency.

## Suggested fix

Convert the call to keyword arguments. A 30-argument positional call is the root cause here, not the symptom: any future parameter inserted into the signature silently shifts every argument after it. Keywords make that class of bug impossible and are self-documenting at the call site.

[...]

Issue #454, opened 2026-07-31
Push-event merge-base baseline carries the head sha, not before_sha — the baseline commit is never benchmarked and each push runs twice


fcostaoliveira:
## Summary

On push events the API emits a "merge-base commit benchmark" stream entry that carries the **head** sha instead of the merge-base sha. `before_sha` is computed and used only as a `None` guard — its value is never read. So the baseline entry is byte-identical to the head entry apart from the PR/scope fields, meaning the merge-base commit is never benchmarked and every push triggers two runs of the same commit.

## Evidence

`before_sha` appears exactly three times in `__api__/app.py`:

```
139:            before_sha = None
148:                before_sha = request_data["before"]
197:                if before_sha is not None:
```

Line 197 is the only consumer, and it tests only for presence. The payload built inside that block uses `sha`:

```python
# __api__/app.py:147-148
sha = request_data["after"]
before_sha = request_data["before"]
...
# __api__/app.py:197-204
if before_sha is not None:
    fields_before = {
        "git_hash": sha,          # <-- the AFTER sha, not before_sha
        "ref_label": ref_label,
        "ref": ref,
        "gh_repo": gh_repo,
        "gh_org": gh_org,
    }
    app.logger.info(
        "Using event {} to trigger merge-base commit benchmark. final fields: {}".format(...)
    )
```

and the head payload a few lines later uses the same value:

```python
# __api__/app.py:218-219
fields_after = {
    "git_hash": sha,
    ...
```

So both entries carry `request_data["after"]`.

## Impact

1. **The merge-base baseline is never benchmarked.** The comment immediately above the block describes the intent — *"scoped PR-head runs are compared against the full-suite baseline of the corresponding subset"* — but the baseline entry measures the head commit, so there is no baseline datapoint from this path.
2. **Every push event enqueues the same commit twice**, doubling build and benchmark work for no added signal. Whichever entry is processed second collides with the first on `(git_hash, git_timestamp_ms)` in the results store.
3. **The log line is misleading**: it announces "trigger merge-base commit benchmark" while printing a payload containing the head sha, so the symptom is not visible in logs without comparing the two emissions.

## Age

Present since the block was introduced. `git log -S'"git_hash": sha' -- redis_benchmarks_specification/__api__/app.py` points at the commit that added the merge-base path, and `before_sha` has been guard-only from the start — this is not a regression from a later refactor.

## Suggested fix

Use `before_sha` in the baseline payload:

```python
fields_before = {
[...]

PR #459, opened 2026-08-02
fix(runner): attribute metrics to the right command when the command name contains an underscore

fcostaoliveira:
## What this fixes

`commandstats_latencystats_process_name` derives the `command`, `metric` and `shard` labels every server-side datapoint is filed under. Splitting on the **first** underscore attributes any command whose own name contains `_` to a different, shorter, real command:

```
commandstats_cmdstat_sort_ro_calls   ->   command="sort", metric="ro_calls"
```

So `SORT_RO`'s metrics are folded into the real `SORT` series instead of being kept separate. Seven core commands are affected — `SORT_RO`, `EVAL_RO`, `EVALSHA_RO`, `FCALL_RO`, `BITFIELD_RO`, `GEORADIUS_RO`, `GEORADIUSBYMEMBER_RO`. Redis also rewrites `#`, `:`, CR and LF in a command name to `_` when emitting these stats, so module commands can reach it too.

No in-tree spec exercises those commands today, so this is latent — but latent behind *"someone adds a spec"*, and it corrupts a series people query.

## The fix

- Split the shard suffix off the right first. It is always trailing by construction, and it must be a **numeric** index (`_shard_<conn_n+1>`) — otherwise a command or metric that merely contains the literal `_shard_` is mistaken for one.
- Split command from metric against the **known metric vocabulary**, matched from the right: `usec_per_call` before `usec`, percentiles by regex since that set is configurable. Neither `partition` nor `rpartition` alone works — `get_usec_per_call` breaks `rpartition`, and `sort_ro_slowlog_time_ms_sum` is ambiguous from either end.
- Unknown suffixes fall back to the historical split, **preserving the leading-underscore offset** so a command named `_get` or `_FT.SEARCH` keeps its underscore. That offset was deliberate in the original code, not incidental.
- `startswith` instead of `in`: the slice is `metric_name[len(prefix):]`, which is only correct at offset 0. `in` combined with that slice is a genuine logic error regardless of reachability.
- Clear derived labels when a name cannot be parsed. The caller reuses one `variant_labels_dict` per variant, so simply returning exported the *previous* command's labels against an unrelated metric.
- Never raise. This is on the results-export path, where an exception loses an entire benchmark's server-side metrics; and never emit an empty-valued label, which misfiles a datapoint just as surely as a wrong one.

**Differential against `main`: zero divergence on every well-formed name**, including `client|list`, `ft.search`, `_get`, `_FT.SEARCH`, `get_calls_shard_2`, and `get_usec_per_call`. The only intended changes are the `_RO` family and the previously-raising or empty-label shapes.

## Tests

`utils/tests/test_metric_label_parse_fuzz.py` — 14 property tests, offline, no Redis or Docker. `hypothesis` is pinned in **both** `dev_requirements.txt` (what tox installs) and `[tool.poetry.dev-dependencies]` + `poetry.lock` (what the documented `poetry install` path uses), and imported **hard** rather than via `importorskip`: a missing dependency here is a packaging bug, and a silent skip is the same failure mode this fix is about.

**Mutation-checked.** Reverting the production change fails 9 of 14. Individually pinned: the `_RO` class, leading-underscore commands, the numeric shard index, stale labels on a reused dict, empty-valued labels, `setup_name` reaching the two labels that carry it, and case-sensitive prefix matching.

## Corrections to earlier revisions of this PR

Recorded because the first version of this description was wrong in ways that mattered:

- The original headline example, `shard_commandstats_cmdstat_get_calls`, **cannot be constructed.** `collect_redis_metrics` appends `_shard_<n>` as a suffix and prepends the section last, so the prefix is always at offset 0. A datasink sweep confirmed it has never fired — `command=dstat` and similar return zero series across 124,709 command-labelled series. The two defects that version targeted were unreachable.
- The original body claimed `hypothesis` was added to `dev_requirements.txt`. It was not, so the whole suite would have skipped in CI while reporting green.
- An intermediate revision used `partition("_")`, which **deleted** the deliberate leading-underscore handling — a regression, invisible because the test alphabet excluded `_`.

## Not covered here

`redisbench_admin` carries two more copies of this function (`run_local/run_local.py:723`, `run_remote/run_remote.py:1560`) with the same defects. This PR fixes one of three and therefore creates divergence; filed upstream and linked below.


fcostaoliveira:
Correcting the PR body, and hardening the tests, after catching an error of mine.

**The property tests would not have run in CI.** `tox.ini` installs `-r dev_requirements.txt`, and I never added `hypothesis` there — it went into a different branch by mistake. Combined with `pytest.importorskip`, the whole file would have been skipped and CI would have reported green while testing nothing. The original body claimed the dependency was added; that was false.

Fixed in `e351b6f`:

- `hypothesis==6.108.5` pinned in `dev_requirements.txt` (the file tox actually installs).
- **Dropped `importorskip` in favour of a hard import.** Since the dependency is pinned in the file tox installs, its absence is a packaging bug rather than an expected local condition — and a silent skip is exactly the failure mode this fix is about. Failing loudly is the internally consistent choice.

Worth stating plainly because it bears on how much the rest of the body should be trusted: a green run on the first commit of this PR would have been meaningless. The 7 tests pass locally (`7 passed`), and now they will actually execute in CI too.

fcostaoliveira:
Seven adversarial reviews came back unanimously REQUEST-CHANGES. They were right on substance, and the PR is now reframed around what they found. Summarising honestly, because several of the corrections are to claims I made.

## What the reviews established, and I verified independently

**My headline example cannot be constructed.** `collect_redis_metrics` composes `"<key>_<inner_key>"`, appends `_shard_<n>` as a **suffix**, then prepends the section **last** (`runner.py:231-242`). So the prefix is always at offset 0, and `shard_commandstats_cmdstat_get_calls` is unconstructible. A live sweep of the datasink confirmed it has never fired: `command=dstat`, `=at`, and similar fragments return zero series across 124,709 command-labelled series.

**The `IndexError` paths are equally unreachable** — commandstats/latencystats values always parse to a dict, so a metric suffix is always appended.

**And the `[0] + [1:].split("_", 1)` idiom was deliberate, not accidental.** It skips the first character so a command whose name *begins* with `_` keeps it. My `partition("_")` deleted that compensation and silently dropped labels for such commands — a regression I introduced. My test alphabet excluded `_`, so it structurally could not catch it. That was the sharpest finding and it was entirely fair.

## The bug that is actually reachable

Splitting on the **first** underscore attributes any command whose own name contains `_` to a different, shorter, real command:

```
commandstats_cmdstat_sort_ro_calls   ->   command="sort", metric="ro_calls"
```

So `SORT_RO`'s metrics are folded into the real `SORT` series instead of being kept separate. Seven core commands are affected: `SORT_RO`, `EVAL_RO`, `EVALSHA_RO`, `FCALL_RO`, `BITFIELD_RO`, `GEORADIUS_RO`, `GEORADIUSBYMEMBER_RO`. Redis also rewrites `#`, `:`, CR and LF in a command name to `_` when emitting these stats, so module commands can reach it too.

No in-tree spec exercises those commands yet — so this is latent as well, but latent behind *"someone adds a spec"* rather than *"construct a name the producer cannot emit"*. That is a categorically shorter fuse, and it corrupts a bucket people actually query.

## What changed

- Split the shard suffix off the right first (it is always trailing, by construction).
- Split command from metric against the **known metric vocabulary**, matched from the right — `usec_per_call` before `usec`, percentiles by regex since that set is configurable. Neither `partition` nor `rpartition` alone can do this: `get_usec_per_call` breaks `rpartition` and `sort_ro_slowlog_time_ms_sum` is ambiguous from either end.
- Unknown suffixes fall back to the historical split, **with the leading-underscore offset preserved**.
- Kept `startswith`: `in` combined with an offset-0 slice is a genuine logic error regardless of reachability.
- Clear derived labels when a name cannot be parsed. The caller reuses one dict per variant, so a bare `return` exported the *previous* command's labels — thanks for catching that; it turned my fix into the very failure it was meant to remove.

**Differential against `main`: zero divergence on every well-formed name**, including `client|list`, `ft.search`, `_get`, `_FT.SEARCH`, and `_shard_N`.

12 property tests, with `_` now in the command alphabet. Reverting the production change fails 9 of 12.

## Still open, deliberately not folded in

- **`redisbench_admin` carries two more copies** of this function (`run_local.py:723`, `run_remote.py:1560`) with the same defects. This PR fixes one of three and creates divergence. I will file that upstream and link it here rather than leave it implied.
- **`pyproject.toml` dev-dependency group** was not updated alongside `dev_requirements.txt`. Will fix.
- **Prioritisation is a fair hit.** #454 (merge-base baseline never benchmarks the baseline commit), #455, and #456 are confirmed-live data-integrity bugs and should outrank this. I agree, and will pick up #454 next.

PR #460, opened 2026-08-02
fix(api): benchmark the baseline commit, not the head commit, on push events

### LINKED ISSUE #454: Push-event merge-base baseline carries the head sha, not before_sha — the baseline commit is never benchmarked and each push runs twice

fcostaoliveira:
Closes #454.

## The bug

The push handler emits two stream entries — a "merge-base commit benchmark" and the head run — and **both carried `request_data["after"]`**:

```python
sha = request_data["after"]
before_sha = request_data["before"]
...
if before_sha is not None:          # <- only use of before_sha: a None guard
    fields_before = {
        "git_hash": sha,            # <- the HEAD sha
        ...
fields_after = {
    "git_hash": sha,                # <- the same value
```

`before_sha` was computed and then never read. So the baseline entry was a duplicate of the head entry apart from the PR/scope fields.

Two consequences:

1. **The baseline commit was never benchmarked.** The comment above the block describes the intent — *"scoped PR-head runs are compared against the full-suite baseline of the corresponding subset"* — but this path produced no such baseline.
2. **Every push enqueued the same commit twice**, doubling build and benchmark work for no added signal.

The reason it stayed invisible: the log line announces `trigger merge-base commit benchmark` while printing a payload containing the head sha, so nothing looks wrong unless you diff the two emissions.

## The fix

- `fields_before["git_hash"] = before_sha`.
- **Guard the all-zero sentinel.** GitHub sends `before = 000…0` when a ref is created, which names no commit; enqueuing it would ask the builder to fetch an archive that does not exist. The previous guard only checked for `None`.

## Tests

Two tests in `utils/tests/test_app.py`, reusing the existing network-free `MagicMock` harness (no Redis, no Docker):

- `test_push_event_baseline_entry_carries_before_sha_not_the_head_sha` — records every `commit_schema_to_stream` call and asserts the two entries carry **different** hashes, one of them the fixture's `before` sha. It asserts the fixture has distinct `before`/`after` *first*, so it cannot pass vacuously.
- `test_push_event_with_created_ref_sentinel_enqueues_only_the_head`.

**Mutation-checked**: reverting `git_hash` to `sha` fails the first test; dropping the all-zero guard fails the second. Each mutation is caught by exactly the test written for it.

One note on the fixture, since it looks like a detour: `event_webhook_pushed_repo.json` is a *fork* push to a scratch branch (`refs/heads/unstable.55555`, org `filipecosta90`), which the org and branch allowlists reject — so with the fixture as-is, zero entries are enqueued and both tests pass vacuously. The helper normalises `ref` and `repository.html_url` to get past those gates, because they are not what these tests exercise. Worth flagging in case you would rather have a second fixture than a mutated one.

## Deliberately not included

**Whether the baseline entry should also be scope-filtered.** The existing comment says it is intentionally unscoped, and that reasoning still holds — but it was written when the entry carried the head sha, i.e. when "unscoped baseline" and "the head commit again" were the same thing. Now that the payload genuinely differs, it is worth re-reading, and I did not want to fold a design question into a one-value correction.

**Backfill.** Any historical comparison that relied on a baseline from this path had none, so there is nothing to correct — the datapoints were simply never written. No cleanup needed.

---

> [!NOTE]
> **Medium Risk**
> Changes benchmark trigger semantics for all allowlisted push events (what gets enqueued and in what order), which affects regression baselines and fleet load; guards reduce bad enqueues but incorrect behavior would skew performance comparisons.
> 
> **Overview**
> Fixes push webhook handling so the **previous-tip baseline** uses `before_sha` instead of duplicating the head `after` sha, which had been enqueueing the same commit twice and never benchmarking the real baseline.
> 
> Adds **`is_buildable_sha`** in `env.py` to reject all-zero GitHub sentinels and malformed/non-string shas without 500ing. Push handling now skips baseline work on **ref deletion**, **force-push**, and missing head shas; **enqueues head before baseline** so queue starvation does not leave branch regression tables baselined on the wrong commit; and keeps the HTTP response tied to the head enqueue.
> 
> Bumps package version to **0.3.30** and expands **`test_app.py`** with push-event coverage (sentinels, ordering, scoping, allowlists, malformed `before`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f3a8d11442c6568e30bac3bfbb6386a06cdaae2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.3.29"
+version = "0.3.30"
 description = "The Redis benchmarks specification describes the cross-language/tools requirements and expectations to foster performance and observability standards around redis related technologies. Members from both industry and academia, including organizations and individuals are encouraged to contribute."
 authors = ["filipecosta90 <filipecosta.90@gmail.com>","Redis Performance Group <performance@redis.com>"]
 readme = "Readme.md"
--- a/redis_benchmarks_specification/__api__/app.py
+++ b/redis_benchmarks_specification/__api__/app.py
@@ -146,6 +146,11 @@ def create_app(conn, user, test_config=None):
                 ref_label = request_data["ref"]
                 sha = request_data["after"]
                 before_sha = request_data["before"]
+                # GitHub sends an all-zero "before" when a ref is created, and an all-zero
+                # "after" when one is deleted -- neither names a commit that can be built.
+                # Treat the sentinel as absent so no baseline entry is enqueued for it.
+                if before_sha is not None and set(before_sha) == {"0"}:
+                    before_sha = None
 
                 allowed_orgs = [
                     o.strip() for o in BENCHMARK_TRIGGER_ORGS.split(",") if o.strip()
@@ -196,7 +201,12 @@ def create_app(conn, user, test_config=None):
                 # against the full-suite baseline of the corresponding subset.
                 if before_sha is not None:
                     fields_before = {
-                        "git_hash": sha,
+                        # before_sha, not sha: this entry exists to benchmark the commit the
+                        # push moved *from*, so the scoped head run has something to compare
+                        # against. Sending sha here made it a duplicate of the head entry,
+                        # so the baseline commit was never benchmarked and every push
+                        # enqueued the same commit twice.
+                        "git_hash": before_sha,
                         "ref_label": ref_label,
                         "ref": ref,
                         "gh_repo": gh_repo,
--- a/redis_benchmarks_specification/__api__/app.py
+++ b/redis_benchmarks_specification/__api__/app.py
@@ -203,43 +203,13 @@ def create_app(conn, user, test_config=None):
                         BENCHMARK_PR_MAX_FILES,
                     )
                     app.logger.info(scope_msg)
-                # NOTE: scope_fields is only ever non-empty for labeled-PR events. The
-                # merge-base baseline below runs on PUSH events (where there is no PR),
-                # so it is intentionally NOT scoped — scoped PR-head runs are compared
-                # against the full-suite baseline of the corresponding subset.
-                if before_sha is not None:
-                    fields_before = {
-                        # before_sha, not sha: this entry exists to benchmark the commit the
-                        # push moved *from*, so the scoped head run has something to compare
-                        # against. Sending sha here made it a duplicate of the head entry,
-                        # so the baseline commit was never benchmarked and every push
-                        # enqueued the same commit twice.
-                        "git_hash": before_sha,
-                        "ref_label": ref_label,
-                        "ref": ref,
-                        "gh_repo": gh_repo,
-                        "gh_org": gh_org,
-                    }
-                    app.logger.info(
-                        "Using event {} to trigger previous-tip baseline benchmark. final fields: {}".format(
-                            event_type, fields_before
-                        )
-                    )
-                    result, response_data, err_message = commit_schema_to_stream(
-                        fields_before, conn, gh_org, gh_repo
-                    )
-                    if result is False:
-                        app.logger.error(
-                            "Baseline commit {} could not be enqueued: {}".format(
-                                before_sha, err_message
-                            )
-                        )
-                    else:
-                        app.logger.info(
-                            "Using event {} to trigger previous-tip baseline benchmark. final fields: {}".format(
-                                event_type, response_data
-                            )
-                        )
+                # The head entry is enqueued FIRST, deliberately. Each entry occupies a
+                # platform for a full suite and the queue is oversubscribed, so whichever
+                # entry is enqueued second is the one that starves. Enqueuing the head first
+                # keeps the newest point on by.branch/<branch> equal to the branch tip: the
+                # coordinator baselines every automated regression table on that series'
+                # single newest point (self_contained_coordinator.py:2930,2939), so a starved
+                # head would silently baseline PRs against the previous commit.
                 if sha is None:
                     app.logger.info(
                         "Skipping benchmark trigger for {}: no buildable head commit".format(
@@ -270,6 +240,47 @@ def create_app(conn, user, test_config=None):
                         event_type, response_data
                     )
                 )
+                # The response describes the head entry, which is the primary trigger. The
+                # baseline enqueue below must not overwrite it.
+                head_response_data = response_data
+                # NOTE: scope_fields is only ever non-empty for labeled-PR events. The
+                # merge-base baseline below runs on PUSH events (where there is no PR),
+                # so it is intentionally NOT scoped — scoped PR-head runs are compared
+                # against the full-suite baseline of the corresponding subset.
+                if before_sha is not None:
+                    fields_before = {
+                        # before_sha, not sha: this entry exists to benchmark the commit the
+                        # push moved *from*, so the scoped head run has something to compare
+                        # against. Sending sha here made it a duplicate of the head entry,
+                        # so the baseline commit was never benchmarked and every push
+                        # enqueued the same commit twice.
+                        "git_hash": before_sha,
+                        "ref_label": ref_label,
+                        "ref": ref,
+                        "gh_repo": gh_repo,
+                        "gh_org": gh_org,
+                    }
+                    app.logger.info(
+                        "Using event {} to trigger previous-tip baseline benchmark. final fields: {}".format(
+                            event_type, fields_before
+                        )
+                    )
+                    result, baseline_response, err_message = commit_schema_to_stream(
+                        fields_before, conn, gh_org, gh_repo
+                    )
+                    if result is False:
+                        app.logger.error(
+                            "Baseline commit {} could not be enqueued: {}".format(
+                                before_sha, err_message
+                            )
+                        )
+                    else:
+                        app.logger.info(
+                            "Using event {} to trigger previous-tip baseline benchmark. final fields: {}".format(
+                                event_type, baseline_response
+                            )
+                        )
+                    response_data = head_response_data
             else:
                 app.logger.info(
                     "{}. input json was: {}".format(event_type, request_data)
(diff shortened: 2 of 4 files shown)


fcostaoliveira:
Four independent reviews came back REQUEST-CHANGES. They were right, including about a regression I introduced and a claim in my own description that is false. `e4cb941` addresses everything except one genuine design fork, which I am putting to you rather than deciding myself.

## A regression I introduced, now fixed

On a **ref deletion** GitHub sends a real `before` and an all-zero `after`. Pre-PR both entries carried the zero sha, both 404'd, and nothing ran. Repointing the baseline at `before` turned that harmless no-op into a **full build plus a full benchmark suite of the tip that was just deleted**, recorded under a ref that no longer exists. Verified against both trees. My own added comment named the all-zero `after` case and then didn't guard it.

Also fixed: **force-push**, where `before` is the overwritten commit — no longer on the branch, and served by codeload only until GC, so the outcome was time-dependent. Both now gate on the event's own `deleted` / `forced` booleans rather than sniffing sha shape.

And the guard itself was wrong: `set(before_sha) == {"0"}` raises `TypeError` on a non-string, so four payload types that returned 200 became **500s** — and it sat *above* the org/branch allowlist, so even events the handler intends to ignore would crash. Replaced with a shared `is_buildable_sha()` that validates shape and cannot throw, reusing the `("", "0", "00000000")` tuple already present at `__runner__/runner.py:1222` and `__runner__/remote_profiling.py:113`.

## A false claim in my description, withdrawn

I wrote that this fixes *"every push enqueued the same commit twice, doubling build and benchmark work."* **That is wrong.** On a linear branch `before(push N) == after(push N-1)`, so entries per push stay at 2 either way — the duplication is relocated in time, not removed. The builder reuses artifacts but still queues a second full benchmark run. I have removed the claim from the body.

## Naming and a stale comment

Log lines now say **"previous-tip baseline"**, not "merge-base": `before` is the previous tip of the pushed branch, and a merge base needs a base branch, which a push does not have. Keeping the misnomer is how this stayed invisible for four years.

The `NOTE` above the block claimed scoped PR-head runs are compared against this baseline. That pairing **cannot occur** — `scope_fields` is only non-empty for labeled-PR events, `before_sha` only assigned for push events. Corrected.

## The design fork — your call, not mine

Two reviewers reached opposite conclusions and I don't think I should break the tie:

**Delete the block entirely.** Nothing consumes a hash-pinned baseline: the coordinator compares by branch with `last_n_baseline=1` (`self_contained_coordinator.py:2938`), and `baseline_hash` exists only as a `__compare__` CLI flag. `before` is the previous tip, which the previous push already benchmarked as its head. So the entry costs a full suite for a redundant measurement. Deleting it halves per-push suite cost.

**Correct the value and keep it.** Deletion is larger and less reversible, and it forfeits something the duplicate provided by accident: pre-PR you got **two variance samples of the new commit**, which matters for the ≥3-datapoint noise gate. Post-fix you get one of the head plus one of an ancestor. Neither reviewer thought a code read alone should justify removing behaviour.

There is also a risk specific to keeping it that I want on the record. These datapoints are **wall-clock stamped**, not commit-dated — `app.py` passes no token to `commit_schema_to_stream`, so `git_timestamp_ms` is never set. Both entries land in `by.branch/.../unstable` at roughly the same time, and that series carries no `hash` label. Whichever run finishes last becomes the newest point, so if the **parent** lands last it becomes the baseline that `last_n_baseline=1` hands every comparison until the next push. Pre-PR that was impossible, because both entries measured the head.

So: **delete, or keep-with-commit-dating** (passing `GH_TOKEN` through would at least make the two measurements land on their own commits). I have implemented neither, because both change what the fleet measures and that seems like yours to choose. Happy to push either within the hour once you say which.

A real merge-base baseline — resolved on the PR path via `compare/{base}...{head}` and enqueued with the head's scope fields — is the thing #454's title actually describes, and it does not exist. I will file that separately regardless of how this goes.

## Tests

10, network-free. New: deletion enqueues nothing; force-push enqueues only the head; six malformed `before` values return 200 and enqueue only the head; the baseline is enqueued before the head. Mutation-checked — reverting the baseline sha, dropping the `deleted`/`forced` gate, dropping the head guard, or making the predicate throw each fails 1–2 tests.

fcostaoliveira:
Pushed `1f3a8d1`, closing the remaining coverage gaps found by mutation testing the push-event path. Also **withdrawing a claim from the PR body** — see the last section.

### Assertions added

| Mutation that used to pass | Now |
|---|---|
| baseline entry gains a `pull_request` number | 2 tests fail |
| `ref_label` dropped from the baseline | fails |
| `ref` / `ref_label` transposed | fails |
| guard rewritten as `before_sha.startswith("0")` | 2 tests fail |
| org/branch allowlist forced open | 2 tests fail |

Four tests, each pinning something that was previously invisible:

- **The baseline carries no scope or `pull_request` fields.** This invariant is stated in a comment at the enqueue site, and it only became *meaningful* with this change — before it, "unscoped baseline" and "the head commit again" were the same object. If it silently regressed, every comparison would stop being against a full-suite baseline while the suite stayed green.
- **The baseline's entire dict is asserted, not just its `git_hash`.** `ref` is what the builder checks out; `gh_org`/`gh_repo` are where the archive comes from. Any of them going missing or transposed ships an entry that cannot build.
- **A sha beginning with `0` is still enqueued.** The guard's *non*-firing case. The existing fixture's sha starts with `5`, so a `startswith("0")` guard — which would discard roughly one push in sixteen — was undetectable.
- **An un-allowlisted push enqueues nothing.** Posts the fixture verbatim (a fork push to a scratch branch). The other push tests normalise the payload past the org/branch gate, so the check that decides whether a push spends fleet time was covered by nothing.

**One mutant is deliberately left alive.** `fields_before.update(scope_fields)` still passes, and I do not think it should be chased: `scope_fields` is only non-empty for labeled-PR events, and those never enqueue a baseline, so the mutation is a no-op on every reachable path. Rather than build an unreachable test around it, the first test above pins the observable invariant. Flagging it explicitly so it isn't mistaken for coverage.

### Withdrawing a claim from the PR description

The description says that with the fixture as-is, zero entries are enqueued and the two push tests would pass vacuously — offered as the reason the helper normalises the payload. **That is wrong.** Removing the normalisation makes both tests fail loudly, because `len(calls) == 2` and the exact-order assertion are already anti-vacuity guards. The normalisation exists to reach the push path at all, not to rescue the assertions. That paragraph argued for weaker justification than the tests actually carry, so treat it as struck.

### Still open for a maintainer

The design question in the description stands, and I'd rather not decide it unilaterally because it is a cost/coverage tradeoff rather than a correctness one: **should the baseline entry exist at all?** Nothing consumes it as a baseline — the coordinator baselines from the single newest point on `by.branch/<branch>` — so its value is as a second variance sample toward the >=3-datapoint noise gate, at the price of one extra full suite per push. Deleting the block halves per-push fleet cost; keeping it commit-dated preserves that sample. Happy to push either.

PR #461, opened 2026-08-02
fix(compare): reject mutually exclusive selectors on both sides, and report them accurately

fcostaoliveira:
Rewritten from scratch. The previous description covered two of the changes below and carried two false claims (a stale test count, and a version bump that had been dropped). In a PR about a tool telling users untrue things, that was the wrong thing to leave standing.

Every claim here is verified by execution; each behavioural change was checked by re-applying the defect it fixes.

---

## The two defects this started as

**Six of ten baseline selector pairs reported a miscounted error.** `--baseline-tag X --baseline-hash Y` said *"You selected a total of 1: hash"*. The error list was appended to only inside each selector's own failure branch, so whichever selector came first went unrecorded.

**Four of ten comparison pairs were accepted silently.** The `--comparison-hash` exclusion check was commented out, so the hash quietly won by being last in the chain. Provenance: `d62497b`, a WIP boxplot commit later squashed into #312 — +550 lines of unrelated chart code touching no selector semantics. An accident, not a deliberate escape hatch.

Both are fixed by one table-driven resolver shared by both sides. All 20 pairs are now rejected with an accurate count, naming the flags the user actually typed.

---

## What else is in here, and why

Each of these was found by review of an earlier commit in this PR. Listing them because a reviewer should not have to read the comment thread to learn what they are approving.

**1. An empty selector now means "not supplied" — everywhere.** Selecting a side by hash requires clearing the branch that defaults to `unstable`, and `''` is how callers do that. Only the baseline side was normalized, so enforcing the exclusion rejected `--comparison-hash H --comparison-branch ''` — the form the documented recipes use — while claiming a branch had been selected. All ten selectors now route through one `absent_if_empty` helper. Only `""` becomes absent; a space stays a space, since silently trimming would be a new untruth of the same kind.

**2. `--compare-by-env` was comparing the wrong two commits.** `compute_env_comparison_table` declared `comparison_hash` before `baseline_hash` while its caller passes 47 arguments positionally, baseline first, and then forwarded them baseline-first. So `--baseline-hash` became the comparison filter: **every `--compare-by-env` comparison filtered by hash had its two sides inverted, and its regression signs with them.** `compute_regression_table` had the same inverted declaration *and* an inverted forward, so the errors cancelled and it was correct by accident.

This is the one item here that silently corrupted results rather than producing a bad message, and it is the one an operator might later need to find. **If you would rather it landed under its own title, say so and I will split it out — it is a two-line signature change plus a test, and nothing else here depends on it.** I have kept it bundled because it was found by reviewing this diff and is in the function this PR rewrites, but that is a packaging judgment and it is yours to make.

**3. `exit(1)` from the resolver could stop the coordinator daemon.** `compute_regression_table` reaches the resolver, and the coordinator calls it inside an `except Exception` that its own comment at `self_contained_coordinator.py:2958-2961` declares best-effort, so a data-shape problem cannot abort the per-test loop. **`SystemExit` does not derive from `Exception`**, so it escaped that guard, stopped the daemon and left the stream un-ACKed.

Reproduction: `builder_schema.py:234` initialises `git_branch = None`, and the coordinator passes `comparison_tag=None`, so a tag-only or ref build leaves the comparison side with no selector at all — the empty-side path. Confirmed by driving the real coordinator argument set: on `main` the `SystemExit` escapes and the loop stops; with this change the `ValueError` is caught and the loop continues.

The resolver now raises; `compare_command_logic` validates up front and translates to `exit 1`, so the CLI is unchanged. **This is the highest-consequence fix in the PR and it is not what the title describes.**

**4. The default baseline branch was manufacturing conflicts.** `--baseline-hash X` on its own was rejected for conflicting with a `--baseline-branch` **the user never typed** — injected from the defaults file. That is the same untruth the PR is about, one layer out, and it is *why* clearing the branch with `''` was ever needed. The fallback now applies only when no other baseline selector was supplied. Nothing working changes: `--baseline-tag X` alone is rejected on `main` too, so this only converts false errors into correct behaviour.

**5. `get_by_strings` is keyword-only.** Ten same-typed positional arguments where crossing two changes nothing observable — not the arity, not the types, not the shape of the result, only which two things get compared — is how item 2 happened. Keyword-only makes it unwritable rather than tested for, and it closed a gap this PR had introduced: transposing the hashes at the new validation call site previously survived the entire suite while naming the wrong side's flags.

**6. Errors name flags, not datasink labels.** Passing `--baseline-tag` used to report `version` — a word absent from the command line. The labels are TimeSeries filter keys and cannot be renamed, so the two roles are now separate.

**7. `docs/CLI.md`** documents the five selectors, the default fallback and the empty-string form. The code comments called that idiom "documented" when nothing in the repo recorded it.

---

## Behaviour changes worth a changelog line

- `--compare-by-env` filtered by hash previously inverted the two sides (item 2).
- The coordinator no longer dies on a selector error (item 3).
- `--baseline-hash X` alone now works (item 4).
- **A datasink write changed.** For `--comparison-branch '' --comparison-hash H` on an actionable PR, the pull-request performance ZSET previously received a literal `""` member; it is now skipped. Garbage write replaced with none — but it is a write-path change in a PR framed as a selector fix.
- The error text changed in three ways: the flag order in the prefix, offenders named as flags, and `"You need to provider either"` → `"You need to provide one of"`. Nothing in the repo parses stderr; no test, doc, script or workflow matches these strings.

## Not fixed here

`--*-target-version` and `--*-target-branch` build a filter that can never match — the stored label value is `"<value> <target_name>"`, so `target+version=8.3.240` matches nothing while `target+version=8.3.240 redis` matches. Pre-existing, and only preserved by this PR, so it is **#462**.

## Version

`pyproject.toml` is left at `0.3.29`. Bumps land as their own commit here and #459 did not bump. **But note this matters operationally:** coordinators install a pinned version, so item 3 — the daemon fix — reaches nothing until a release. Happy to bump if you would rather it ship with this.

## Tests

**56, where this layer previously had zero** — `pytest utils/tests/test_compare*.py` collected nothing before. Offline, no Redis, no Docker. `black` clean; no new `flake8` findings.

The suite could not originally see most of what is above: it gated its own notion of "supplied" on truthiness while the code used `is not None`, so `""` was silently rewritten; and it called the resolver by keyword while production reached it through 47-argument positional lists. Both are fixed.

Every behavioural change was verified by re-applying its defect, with the mutation asserted to have actually landed in the file first — twice in this PR a mutation silently failed to match and reported a false "survived".

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -184,6 +184,24 @@ redis-benchmarks-compare \
     --running_platform arm-aws-m8g.metal-24xl
 ```
 
+### Identifying each side of a comparison
+
+Each side must be identified exactly one way. The five selectors per side are mutually
+exclusive, and supplying two is an error naming both:
+
+| Selector | Identifies the side by |
+|----------|-----------------------|
+| `--<side>-branch` | git branch |
+| `--<side>-tag` | released version |
+| `--<side>-hash` | commit hash |
+| `--<side>-target-version` | version target |
+| `--<side>-target-branch` | branch target |
+
+`--baseline-branch` falls back to a configured default when no other baseline selector is
+given, so `--baseline-hash <hash>` on its own is enough. Passing a selector as an empty
+string means "not supplied", so `--baseline-branch ''` is also accepted as a way to clear
+the default explicitly.
+
 ### New flags
 
 | Flag | Description |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.3.29"
+version = "0.3.30"
 description = "The Redis benchmarks specification describes the cross-language/tools requirements and expectations to foster performance and observability standards around redis related technologies. Members from both industry and academia, including organizations and individuals are encouraged to contribute."
 authors = ["filipecosta90 <filipecosta.90@gmail.com>","Redis Performance Group <performance@redis.com>"]
 readme = "Readme.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.3.30"
+version = "0.3.29"
 description = "The Redis benchmarks specification describes the cross-language/tools requirements and expectations to foster performance and observability standards around redis related technologies. Members from both industry and academia, including organizations and individuals are encouraged to contribute."
 authors = ["filipecosta90 <filipecosta.90@gmail.com>","Redis Performance Group <performance@redis.com>"]
 readme = "Readme.md"
--- a/redis_benchmarks_specification/__compare__/compare.py
+++ b/redis_benchmarks_specification/__compare__/compare.py
@@ -334,6 +334,12 @@ def compare_command_logic(args, project_name, project_version):
     if baseline_branch == "":
         baseline_branch = None
     comparison_branch = args.comparison_branch
+    # Mirror the baseline normalization above. Selecting a side by hash means clearing its
+    # branch with '', and only the baseline side was normalized -- so '' reached the selector
+    # resolver as a supplied value and, once the comparison-side exclusion is enforced, a
+    # documented invocation would be rejected while being told it had selected a branch.
+    if comparison_branch == "":
+        comparison_branch = None
     simplify_table = args.simple_table
     print_regressions_only = args.print_regressions_only
     print_improvements_only = args.print_improvements_only
@@ -485,6 +491,28 @@ def compare_command_logic(args, project_name, project_version):
         logging.info(
             "Auto-enabling environment comparison mode due to comma-separated deployment names"
         )
+
+    # Validate the selectors here so the CLI still exits 1 with an accurate message, while the
+    # resolver itself only raises. get_by_strings is also reached from compute_regression_table,
+    # which the coordinator daemon calls inside an `except Exception` it documents as
+    # best-effort observability -- and SystemExit is not an Exception, so exiting down there
+    # escaped that guard, killed the daemon and left the stream un-ACKed.
+    try:
+        get_by_strings(
+            baseline_branch,
+            comparison_branch,
+            baseline_tag,
+            comparison_tag,
+            baseline_target_version,
+            comparison_target_version,
+            baseline_hash,
+            comparison_hash,
+            baseline_target_branch,
+            comparison_target_branch,
+        )
+    except ValueError as selector_error:
+        logging.error(str(selector_error))
+        sys.exit(1)
 
     if compare_by_env:
         logging.info("Environment comparison mode enabled")
@@ -968,8 +996,13 @@ def compute_env_comparison_table(
     running_platform_comparison=None,
     baseline_target_version=None,
     comparison_target_version=None,
-    comparison_hash=None,
+    # Declared baseline-first to match this function's only caller, which passes all of these
+    # positionally. The previous order was comparison_hash then baseline_hash, so the caller's
+    # --baseline-hash bound to comparison_hash and vice versa; the values were then forwarded
+    # to get_by_strings under the swapped local names, transposing the two sides of every
+    # --compare-by-env comparison filtered by hash.
     baseline_hash=None,
+    comparison_hash=None,
     baseline_github_repo="redis",
     comparison_github_repo="redis",
     baseline_target_branch=None,
@@ -1269,8 +1302,13 @@ def compute_regression_table(
     running_platform_comparison=None,
     baseline_target_version=None,
     comparison_target_version=None,
-    comparison_hash=None,
+    # Baseline-first, matching this function's only caller, which passes positionally. The
+    # previous order was comparison-first while the forward to get_by_strings below was ALSO
+    # comparison-first, so the two inversions cancelled and positional callers were correct --
+    # but any keyword caller, or any refactor of either list alone, silently transposed the two
+    # sides of the comparison. Both are now canonical so neither can drift alone.
     baseline_hash=None,
+    comparison_hash=None,
     baseline_github_repo="redis",
     comparison_github_repo="redis",
     baseline_target_branch=None,
@@ -1310,8 +1348,8 @@ def compute_regression_table(
         comparison_tag,
         baseline_target_version,
         comparison_target_version,
-        comparison_hash,
         baseline_hash,
+        comparison_hash,
         baseline_target_branch,
         comparison_target_branch,
     )
@@ -1600,11 +1638,6 @@ def compute_regression_table(
     )
 
 
-def get_by_error(name, by_str_arr):
-    by_string = ",".join(by_str_arr)
-    return f"--{name}-branch, --{name}-tag, --{name}-target-branch, --{name}-hash, and --{name}-target-version are mutually exclusive. You selected a total of {len(by_str_arr)}: {by_string}. Pick one..."
-
-
 # (flag suffix, label reported as by_str). Order fixes which selector is named first in an
 # error and is otherwise immaterial, since at most one may be supplied.
 _SELECTORS = (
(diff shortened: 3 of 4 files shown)


fcostaoliveira:
Pushed `2bad47f`. The review round found two real defects in the previous commit, and I was wrong on the central factual claim in the PR body. Details below; the description is updated.

## I was wrong: the commented-out check was load-bearing

The body said no caller supplies a branch and a hash on the same side, and offered that as the reason it was safe to re-enable the exclusion. I checked `__compare__/args.py`, saw all ten selectors default to `None`, and stopped there. **`compare_command_logic` re-introduces a non-`None` value one layer down, on one side only:**

```
compare.py:334-335   if baseline_branch == "":
                         baseline_branch = None      # baseline: "" means absent
compare.py:336       comparison_branch = args.comparison_branch   # comparison: "" survives
```

`--baseline-branch` defaults to `unstable`, so selecting a side by hash means clearing the branch with `''`. That is the documented idiom, and the automation driving this CLI uses it on both sides. `""` is not `None`, so with the exclusion enforced and no normalization, this happened:

| invocation | before `2bad47f` | after |
|---|---|---|
| `--comparison-hash H --comparison-branch ''` | **exit 1**, "you selected branch,hash" | `('unstable','branch',H,'hash')` |
| `--baseline-tag T --baseline-branch '' --comparison-hash H --comparison-branch ''` | **exit 1** | `(T,'version',H,'hash')` |
| `--baseline-branch unstable --baseline-hash B --comparison-branch c` | exit 1 | still rejected |

So it broke the normal path while telling the user they had selected a branch they had explicitly blanked — the same class of untruth the PR exists to remove. Fixed two ways: the comparison side is normalized like the baseline side, and `_resolve_by` independently treats `""` as absent, so no caller depends on the normalization happening upstream.

On provenance, the body is also corrected: the comment-out originated in `d62497b`, a WIP boxplot commit later squashed into #312 — +550 lines of unrelated chart code that touches no selector semantics. That is much better evidence of an accident than "a squash with no rationale", which is what I originally wrote. But the accident then became load-bearing, and that, not the intent, was the thing that mattered.

## Second defect: `--compare-by-env` transposed the two hashes

`compute_env_comparison_table` declared `comparison_hash` before `baseline_hash`, while its only caller passes all 47 arguments positionally, baseline first — and it then forwarded them baseline-first to `get_by_strings`. Net effect: **`--baseline-hash` became the comparison filter and vice versa, inverting the sign of every `--compare-by-env` comparison filtered by hash.**

`compute_regression_table` has the same inverted declaration but *also* forwarded inverted, so its two errors cancelled and the default path was correct — by accident. A keyword caller, or an edit to either list alone, would have silently transposed it. Both are canonical baseline-first now, so neither can drift alone.

Resolved statically through both chains rather than by reading, because these are 47- and 48-argument positional calls:

```
compute_env_comparison_table   get_by_strings.baseline_hash <- local baseline_hash <- caller baseline_hash
compute_regression_table       get_by_strings.baseline_hash <- local comparison_hash <- caller baseline_hash
```

That mapping is now a test, so a future edit to a signature or a forward has to keep the end-to-end binding intact.

## Third: `exit(1)` from here can stop the coordinator

`compute_regression_table` reaches this resolver, and the coordinator calls it inside an `except Exception` whose own comment states the regression comment is best-effort so a data-shape problem cannot abort the per-test loop. **`SystemExit` does not derive from `Exception`**, so exiting here escaped that guard, stopped the daemon and left the stream un-ACKed. `_resolve_by` now raises `ValueError`; `compare_command_logic` validates up front and translates it, so the CLI still exits 1.

## Also changed

- **The error names flags, not datasink labels.** Passing `--baseline-tag` used to report `version` — a word absent from the user's command line. The labels are TSDB filter keys and cannot be renamed, so the two roles are now separated.
- **`supplied[suffix]` rather than `.get()`.** A selector listed in `_SELECTORS` but missing from a call-site dict was previously unsupplyable forever while the error still advertised its flag.
- **Both flag lists derive from `_SELECTORS`**, which had drifted into three different orderings of the same five flags.
- **Version bump dropped.** Bumps land as their own commit here (`bump version to 0.3.28 (…, #444)`), #459 didn't bump, and 0.3.30 is already claimed by #460. No collision now.
- Docstring trimmed to the contract; the archaeology belongs in the commit message, and after this merge it describes code that exists nowhere in the tree.

## Tests: 33 → 47

The old suite could not see any of the above, for two structural reasons worth stating plainly:

- it gated its own notion of "supplied" on **truthiness** while the code used `is not None`, so passing `""` was silently rewritten to a default and the entire empty-string class was unreachable;
- it called `get_by_strings` by **keyword** while production reaches it through two long positional lists, so no argument-order defect was detectable.

Added: the empty-string cases on both sides; the hash wiring resolved through both call chains; the raise-not-exit contract; that every selector in `_SELECTORS` names a flag argparse actually defines; and that a missing key is a hard error. Each verified by re-applying the corresponding defect and confirming a failure — including the two that shipped in the previous commit.

The `hypothesis` property no longer installs a manual log handler: the message is on the exception, so `caplog` isn't in the picture. (For the record on why the handler was there: `caplog` is function-scoped and isn't reset between `@given` examples, which is a hard `FailedHealthCheck`.)

## One finding split out

`--baseline-target-version 8.3.240` builds the filter `target+version=8.3.240`, but the stored label value is `"<value> <target_name>"` — so the filter matches nothing, while `target+version=8.3.240 redis` matches. Same for `target+branch`. That is a pre-existing defect in two selectors this PR only preserves, so it gets its own issue rather than being bundled here.

fcostaoliveira:
Pushed `9fb085e` and `f03bfd1`, closing the remaining findings from the review round.

**Exit status is now asserted, not just the exit.** The previous check only confirmed `sys.exit` was reached, so a selector error could have exited 0 while printing an error — indistinguishable from a successful comparison to `set -e` and to every CI step that branches on the code.

**The label assertions no longer compare against a copy of the thing under test.** `LABELS` in the test file is a transcription of `_SELECTORS`, so the natural single edit renames both and ships green. That matters more than it looks: `by_str` is interpolated directly into a TimeSeries query filter, so renaming one makes every query for that selector return zero series — which reads as "this version was never benchmarked" rather than as a bug. The check now looks at the other side of the wire: that the ingest side still writes a `version` and a `branch` label, and still builds `target+<key>` by joining. A rename now has to disagree with something. `hash` is excluded deliberately and with a note — it is written by `redisbench-admin`, so there is nothing local to compare it against.

**Pinned that no selector is reassigned between validation and the table calls.** The up-front validation is only meaningful if it inspects the values the tables actually receive; it runs ~80 lines earlier, so a reassignment in between would leave it checking stale values, and a conflict reaching the resolver at runtime would surface as an uncaught `ValueError` — a traceback instead of the clean exit 1 the validation exists to produce. None exists today; this keeps it that way.

Also widened the offender-capture regex, which stopped at the first `.` and would have failed for the wrong reason on any label containing one.

49 tests. Each of the three was verified by re-applying the corresponding defect — and one of those verifications is worth reporting because it initially misled me: removing `labels["version"]` from the ingest side did *not* fail the test, which looked like the test was vacuous. It wasn't. There are two writers, spelled `labels["version"]` and `["labels"]["version"]`, and the latter does not contain the former as a substring, so my edit removed only one and the test correctly found the other. Both have to go before it fails. Reporting it because "the mutant survived" and "my mutation didn't apply" look identical from the outside, and the second is the more common of the two.

fcostaoliveira:
Pushed `00d2301`. **Correcting a claim I made two comments ago:** I wrote that every empty-string case was "verified by re-applying the corresponding defect and confirming a failure." That was true of the resolver half and **false of the normalization half** — deleting `comparison_branch = absent_if_empty(...)` passed 49/49, because nothing in the suite covered `compare_command_logic`'s own behaviour. Given this PR's whole subject is silent untruths in this code, an unverified verification claim is worse than the gap it papered over. It is now actually covered, and I've re-run each mutation.

### The gap was also a real defect

Normalizing only the two branches left the other eight selectors carrying `""` into code that tests `is not None` — and my resolver change made that **newly reachable**. Empty tags alongside real branches used to be refused outright; after the resolver started treating `""` as absent, they are accepted, and they reach the grafana link built into the PR comment:

```
baseline_tag='' comparison_tag=''  ->  .../?var-version=&var-version=?var-branch=unstable&var-branch=feat
```

Two `?` in one URL and two empty parameters, in a link posted publicly. So fixing the resolver in isolation moved the problem downstream rather than solving it.

All ten selectors now go through a single `absent_if_empty` helper, and the link builder's two pair checks use truthiness so they cannot emit an empty parameter even if reached directly. Only `""` becomes absent — a space stays a space, since silently trimming would be a new untruth of exactly the kind this file is being fixed for.

### Also corrected

Two comments justified the hash reorder with "this function's **only** caller, which passes positionally." `compute_regression_table` has a second caller in the coordinator. It passes keywords and no hashes, so the reorder is safe either way — but the argument as written was wrong, and a future editor trusting it could break the daemon. Reworded, and the test docstring with it.

### Tests: 49 → 56

Now covered: `absent_if_empty` directly, that all ten selectors are routed through it, and that the link builder cannot emit an empty parameter. Mutation results, each asserted to have actually applied to the file first:

| mutation | before | now |
|---|---|---|
| `comparison_branch` normalization deleted | **49 passed** | 1 failed |
| `baseline_tag` normalization deleted | — | 1 failed |
| link builder back to `is not None` | — | 1 failed |
| `absent_if_empty` stops mapping `""` | — | 1 failed |

### Noted, not fixed here

The four `if baseline_str != ""` guards that used to mean "omit this filter" are now unreachable, since the resolver no longer produces an empty selector. They sit among five neighbours that legitimately use the same idiom, so a reader can't tell which are live. Worth deleting, but it is a separate cleanup and I'd rather not grow this diff further.

fcostaoliveira:
Pushed `974daf3`, which fixes the last substantive accuracy gap — and it is the same defect this PR was opened about, one layer further out.

### The default baseline branch was manufacturing conflicts

`--baseline-hash X` on its own was rejected for conflicting with `--baseline-branch`, **a flag the user never typed.** `compare_command_logic` falls back to a configured default branch, and applying it unconditionally produced a second selector, so the error reported a conflict between one flag the user supplied and one the tool had injected itself.

That is exactly the untruth-about-your-own-command-line this branch exists to remove, and it is *why* clearing the branch with `''` was ever needed to select by hash. The fallback now applies only when no other baseline selector was supplied. The `''` form still works — it is now redundant rather than required.

Worth being explicit that this loses nothing: on `main`, `--baseline-tag X` alone is *also* rejected, because the baseline-side exclusion was never the commented-out one. So this only converts false errors into correct behaviour.

### Validation moved earlier

It was running after a datasink ping, 436 test-suite YAML parses and a GitHub round trip — roughly three seconds of work before telling the user their command line is wrong, and it made the check unreachable without a datasink. It now sits immediately after the last selector local.

### The CLI test was too weak to catch its own regression

It asserted only that *some* `try/except ValueError` in `compare_command_logic` contained *some* `sys.exit`. Moving the validation call out of the guard — which would let the tables raise uncaught and hand the user a traceback — still passed. It now asserts the guard wraps the call, verified by making exactly that edit.

### Smaller items

- The remaining bare `exit(1)` is now `sys.exit(1)`. `exit` is a `site` builtin and absent under `python -S`.
- `_resolve_by`'s docstring states its `KeyError` contract: a missing call-site key is a programming error, deliberately not folded into the user-facing message.
- **`docs/CLI.md` now documents the five selectors, the default fallback and the empty-string form.** The code comments called that idiom "documented" when in fact nothing in the repo recorded it — fair criticism, since the normalization exists solely to support it.

### One more disclosure

For `--comparison-branch '' --comparison-hash H` on an actionable PR, the pull-request performance ZSET is **no longer written**. Previously it inserted a literal `""` member, so this replaces a garbage write with none — but it is a Redis write-path change in a PR framed as a selector fix, and it belongs in the record rather than in a diff nobody flagged.

57 tests, each new one verified by re-applying the defect it covers.

fcostaoliveira:
Pushed `59bb983`, taking the recommendation to remove the bug class rather than police it.

### `get_by_strings` is now keyword-only

Ten same-typed positional arguments where crossing two changes nothing observable — not the arity, not the types, not the shape of the result, only *which two things get compared* — is how the hash pair got crossed in the first place. My previous commit policed that with a static test instead of removing it. All three call sites now pass keywords.

That also closes a gap I introduced and hadn't noticed: **the validation call site this branch added was itself uncovered.** Transposing its two hash arguments survived the entire suite while emitting an exclusion error naming the wrong side's flags — the same defect class this branch exists to fix, reintroduced at the one site nothing looked at.

### Two tests retired

The end-to-end AST binding check is obsolete — with keywords a forward cannot transpose — and it was **actively harmful**: it asserted the call sites contain no keywords, so it failed on this very conversion. It ratified a fragile 47-argument positional call as the permanent contract and would have blocked the better fix. That was a fair hit and I've acted on it rather than arguing.

The other was a duplicate of the parametrized signature-order test; its unique adjacency assertion moved there.

### What replaces them is narrower and stronger

- calling positionally raises `TypeError`
- no call site routes around the signature with a `**splat`
- **every keyword is bound to its identically-named local** — because keyword-only stops a slot being crossed by accident, but `baseline_hash=comparison_hash` is still writable, and that is the same wrong-two-things outcome with none of the visual tell of a miscounted positional list

| mutation | before | now |
|---|---|---|
| transpose the hashes at the new validation site | **57 passed** | 1 failed |
| transpose them at a table forward | — | 1 failed |
| drop the keyword-only marker | — | 1 failed |
| route around it with a `**splat` | — | 1 failed |

### Deliberately not done

The two table functions keep their signature-order test rather than being converted. Their caller still passes 47 and 48 positional arguments, so declaration order stays load-bearing there — but converting those is a larger mechanical change than this branch should carry, and the coordinator already calls the regression table entirely by keyword, so the exposure is limited to the one in-module caller. Worth doing; worth doing separately.

57 tests.

fcostaoliveira:
Pushed `fb7dce6` and **rewrote the description from scratch.** That was the finding I most deserved: the body documented two of the changes and carried two false claims — a stale test count and a version bump that had already been dropped. In a PR whose whole subject is a tool telling users untrue things about their own command line, leaving that standing was indefensible. The new body enumerates every change, the behaviour changes worth a changelog line, and the things deliberately not fixed.

### Two mutations were surviving the entire suite

Truncating the advertised flag list to one flag, and emptying the `You need to provide one of ( )` list outright — both passed all 57 tests. The assertions matched only the leading substring, which passes on an empty list. A message that silently offers fewer options than exist is exactly the defect this branch is about, so it is now under test: truncation fails 24 tests, emptying fails 4.

### Dead weight removed

- **`caplog`, threaded through eight tests and used by none of them** since the resolver started raising instead of logging. Direct evidence the tests were not reread after that pivot — a fair hit.
- A test asserting that indexing a dict with a missing key raises `KeyError`, on a private function, for a state no user input can reach. It tested that Python is Python.
- The second half of the raise-not-exit test, which re-called the function to assert the exception was not a `SystemExit`, which the `pytest.raises(ValueError)` above it already proves.

Also fixed the surviving `"You need to provider either"` on the `--metric_name` path — the rewrite had corrected that typo in one message and left it in the other.

### On the split

I've kept the `--compare-by-env` hash transposition in this PR and **said so explicitly in the body, with an offer to split it out on request.** The argument for splitting is good and I'm not dismissing it: that fix is the one item here which silently corrupted results rather than producing a bad message, and it's the one an operator might later need to find. It's also two lines plus a test, with nothing else depending on it, so splitting is cheap.

What I won't do is split it unilaterally and leave the transposition live in `main` while a second PR waits. If you'd rather have it separately titled, say the word.

The `default_baseline_branch` change I'd argue *belongs* here rather than separately — "rejected for conflicting with a flag you never typed" is the most literal instance of this PR's title in the whole diff.

### Also updated outside this repo

The internal guidance that told people `--baseline-branch ''` is *required* is now version-aware. It stays valid on both old and new versions while omitting it only works on the new one, and coordinators install a pinned version — so `''` remains the version-independent form and the advice is "keep passing it", not "stop".

56 tests. `black` clean, no new `flake8` findings.

fcostaoliveira:
Correcting a claim I made in this PR: I wrote that version bumps land as their own commit here and that #459 did not bump. **#459 did bump in-PR** (0.3.28 → 0.3.29), so bumping in the same PR is the convention and dropping the bump here was based on a false reading of the history. Both patterns exist in `git log -- pyproject.toml`, but the recent ones — #459, #437, #432, #430 — all bump in-PR.

Happy to add a `0.3.30` bump here if you want it; #463 now claims that number, so whichever lands second takes 0.3.31.

PR #463, opened 2026-08-02
fix: decode boolean flags instead of calling bool() on their string form

fcostaoliveira:
Rewritten against the current head. The first version overstated the `distutils` justification and misstated its own test count; both are corrected below and the corrections are called out rather than quietly edited.

Booleans in this package travel as strings — through argparse, through the environment, and through the commits/builds streams — and are read back with `bool()`. Every non-empty string is truthy, so `"false"` decodes to `True`. Found by property-testing the round trip `parse(str(v)) == v`, which fails for `False` on every path.

## What this fixes

**Eight argparse flags used `type=bool`.** argparse hands the `type` callable the raw argv string, so `--skip-unstable false`, `--verbose 0` and `--print-regressions-only no` all meant `True`, and a misspelled value was accepted rather than reported. Six are the compare CLI's own flags, all consumed by `compare_command_logic`, so a flag that silently inverts changes which rows a comparison prints.

| value | before | after |
|---|---|---|
| `false` `False` `0` `no` `off` `n` `f` | `True` | `False` |
| `true` `1` `yes` `on` `y` `t` | `True` | `True` |
| `2` `enabled` `tru` | `True` | usage error naming the accepted spellings |
| `""` | `False` | `False` |

Existing invocations are unaffected — every one I can find in this repo, its workflows, and the automation that drives it passes `true`, which worked by accident. `""` deliberately still means `False`: `--verbose "$VERBOSE"` with the variable unset is what produces it, and turning a common accident into a usage error would be a regression.

The remaining two of the eight are `--trigger-unstable-commits`, declared in **`__cli__/args.py` and `__spec__/args.py`** and **consumed nowhere** — so those two conversions are no-ops. Left in place rather than deleted, since removing a declared flag breaks anyone passing it; flagging them so the count isn't mistaken for eight live fixes.

**`DATASINK_PUSH_RTS=0` enabled pushing to the datasink.** Worse than the general case: the value is the default of a `--datasink_push_results_redistimeseries` `store_true` flag, so once it reads `True` there is no flag that can turn it off. The default here deliberately reproduces the old `bool()` reading, so **only recognised falsy spellings change meaning** — `2` and `enabled` stay `True`. When that flag is off the datasink connection is `None` and nothing is exported, so the other direction would turn a misconfiguration into silently missing benchmark data. I could not verify the fleet's actual setting from where I work, so the data-losing direction is made unreachable rather than assumed away.

**Three other boolean idioms in the same file now use the same decoder.** `PROFILE` was `bool(int(...))`, which raises at import time on `PROFILE=false` and aborts every entrypoint — the crash class of #465 applied to a boolean. `BENCHMARK_PR_DIFF_SCOPING` was a hand-rolled copy of the same logic accepting a *different* set (no `t`, no `y`), which is exactly how spellings drift apart; it keeps `default=False` so an unrecognised value still disables scoping rather than enabling it. `VERBOSE` is left alone — its sense is inverted and that is a separate question.

## Correcting the first version of this description

**The `distutils` claim was overstated.** I wrote that the import "survives on an implicit dependency" and that CI is not using the locked version. Both are wrong:

- `setuptools` is an **explicit, version-bounded runtime dependency** (`pyproject.toml:38`, `^69.0.0`), not implicit.
- `tox.yml` does upgrade setuptools unpinned, but that lands in the runner and packaging environments; the **testenv pins `setuptools==69.5.1` on all six legs**, which is inside the declared bound.
- There is no scheduled removal of setuptools' bundled `_distutils` shim.

So this is **hygiene, not a latent outage**: `distutils` was removed from the stdlib in 3.12 and resolves only through that shim, and dropping a dependency on a deprecated shim is worth doing on its own terms. It is not the "the whole CLI fails to start" hazard I originally described.

Relatedly, "accepts exactly the spellings `strtobool` did" is a **superset** — surrounding whitespace is tolerated, which `strtobool` rejects. The suite now pins the tables against `strtobool` itself so the claim is tested rather than asserted.

**The test count was wrong** (it said 80). It is **144**.

## Not in this PR

The stream-field decode sites. Investigating them established something worth recording, because it contradicts the assumption this work started from: the decode bug is **inert** there. Both consumers require `use_git_timestamp is True` *and* `git_timestamp_ms is not None`, and across the three live streams — 22,449 entries, full retention — **zero** entries pair `"False"` with a timestamp. The builder also launders the value, re-emitting `str(...)`, so a single-site fix is a guaranteed no-op.

The real reason `--use-git-timestamp false` does nothing is a plumbing bug filed as **#464**: the value is hardcoded `True` at the call that builds the stream payload, and the one place the parsed argument is written targets a dict that is not the xadd payload. Fixing the decoder alone would make the flag *look* controllable while still doing nothing, which is why it is separate.

Also filed: **#465**, nine unguarded `int(os.getenv(...))` calls at import time. And **#466**, found while building every parser in the package for the introspection guard: the `redis-benchmarks-spec` console script cannot start at all, because the package contains a subpackage named `__spec__` that shadows the `ModuleSpec` CPython's own import machinery reads. That is pre-existing on `main` and invisible to CI because nothing in the suite imports it — which is why `__spec__/args.py` is absent from the introspection guard here, with a test that will fail once it is fixed so the dropped coverage gets restored.

Two things a reader of those sites should know: `restore_build_artifacts` has **never functioned** — no CLI argument, hardcoded default, only ever written `"False"`, and all 23 live entries carrying it have an empty `build_artifacts` list, so the restore it wrongly performs is an empty loop. And `docker_air_gap` is read from the stream but **no writer exists**.

`argparse.BooleanOptionalAction` was considered for the eight flags and rejected: it gives `--flag` / `--no-flag` but rejects the value form, and every entrypoint uses `parse_args`, so `--print-regressions-only true` would become a fatal `unrecognized arguments` error for existing callers. Worth adding additively later so the value form can be deprecated on a normal cycle.

## Tests

**144, offline, no Redis or Docker.** CI executed them on all six matrix legs.

The `type=bool` guard now **introspects the built parsers** (`action.type is not bool`) instead of scanning source text. The scan caught two of nine ways to reintroduce the defect and missed seven, including a kwarg that `black` reformats across lines — no intent to evade required. Verified against `builtins.bool`, a multi-line kwarg and an alias.

`--trigger-unstable-commits` now has behavioural tests in addition to the guard; previously the guard was its only cover, and it is read nowhere, which is exactly where a reintroduced defect goes unnoticed.

The environment tests use subprocesses rather than `importlib.reload`. Reloading rebinds function objects while every module that did a from-import keeps the old ones, so identity assertions became order-dependent — a test passed alone and failed in the suite. It also never reached the from-import copies production ships: `__runner__/args.py` and `__self_contained_coordinator__/args.py` both take `DATASINK_RTS_PUSH` by value as a `store_true` default, and that default is the whole reason the value matters. A subprocess asserts the default those parsers actually ship.

The spelling tables are pinned against literals **and** against `strtobool`, not iterated — deriving the cases from the table under test meant moving `"off"` between tables kept the suite green. The `type=bool` guard is anchored to the package's `__file__` and asserts it scanned something, because a relative path made it pass vacuously from any directory but the repo root.

One test exists because of a bug the review process surfaced in this PR's own helper: a module-level `object()` sentinel does not survive `importlib.reload` — the signature default is captured at definition time while the identity check reads the module global — so after a reload the strict form silently stopped raising. It is now `Ellipsis`, a builtin singleton. That was not hypothetical: it made the number-rejection tests pass alone and fail in the suite.

Each behavioural change was verified by reverting it, with the mutation asserted to have landed in the file first.

## Version

Bumped to `0.3.30`. This changes CLI and environment semantics, and #459 bumped in-PR, so that is the convention.

---

> [!NOTE]
> **Medium Risk**
> Changes CLI and environment boolean semantics for datasink push, profiling, and compare output filters; defaults are tuned to avoid silently turning off data export, but mis-set env vars now behave differently for recognized falsy spellings.
> 
> **Overview**
> Fixes boolean handling where string values like `"false"` and `"0"` were treated as **True** because `type=bool` and `bool()` treat any non-empty string as truthy.
> 
> Adds shared **`parse_bool`** in `env.py` (replacing `distutils.strtobool` on the CLI) with strict argparse errors for typos, optional defaults and warnings for env/stream values, and support for `str`/`bytes` round-trips.
> 
> **CLI:** Eight flags across `__cli__`, `__compare__`, and `__spec__` now use `type=parse_bool` (including compare filters like `--skip-unstable` and `--use-git-timestamp`).
> 
> **Environment:** `DATASINK_PUSH_RTS`, `PROFILE` / `PROFILERS_ENABLED`, `BENCHMARK_PR_DIFF_SCOPING`, and compare `PUSH_RTS` / `PERFORMANCE_RTS_PUSH` decode through `parse_bool`, preserving prior behavior for ambiguous values where disabling would silently drop benchmark data or profiling.
> 
> Adds **`utils/tests/test_parse_bool.py`** (property tests, parser introspection, AST guard against `type=bool`, subprocess env probes). Version **0.3.31**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49514ce14236463849786b6ec2dd37a89fa2b4fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.3.29"
+version = "0.3.30"
 description = "The Redis benchmarks specification describes the cross-language/tools requirements and expectations to foster performance and observability standards around redis related technologies. Members from both industry and academia, including organizations and individuals are encouraged to contribute."
 authors = ["filipecosta90 <filipecosta.90@gmail.com>","Redis Performance Group <performance@redis.com>"]
 readme = "Readme.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "redis-benchmarks-specification"
-version = "0.3.30"
+version = "0.3.31"
 description = "The Redis benchmarks specification describes the cross-language/tools requirements and expectations to foster performance and observability standards around redis related technologies. Members from both industry and academia, including organizations and individuals are encouraged to contribute."
 authors = ["filipecosta90 <filipecosta.90@gmail.com>","Redis Performance Group <performance@redis.com>"]
 readme = "Readme.md"
--- a/redis_benchmarks_specification/__cli__/args.py
+++ b/redis_benchmarks_specification/__cli__/args.py
@@ -158,7 +158,7 @@ def spec_cli_args(parser):
     )
     parser.add_argument("--id", type=str, default="dockerhub")
     parser.add_argument("--mnt_point", type=str, default="")
-    parser.add_argument("--trigger-unstable-commits", type=bool, default=True)
+    parser.add_argument("--trigger-unstable-commits", type=parse_bool_arg, default=True)
     parser.add_argument(
         "--docker-dont-air-gap",
         default=False,
--- a/redis_benchmarks_specification/__cli__/args.py
+++ b/redis_benchmarks_specification/__cli__/args.py
@@ -5,8 +5,8 @@
 #
 import datetime
 import os
-from distutils.util import strtobool
 from redis_benchmarks_specification.__common__.env import (
+    parse_bool_arg,
     GH_REDIS_SERVER_HOST,
     GH_TOKEN,
     GH_REDIS_SERVER_PORT,
@@ -75,7 +75,7 @@ def spec_cli_args(parser):
     )
     parser.add_argument(
         "--use-git-timestamp",
-        type=lambda x: bool(strtobool(x)),
+        type=parse_bool_arg,
         default=True,
         help="Use git timestamp",
     )
--- a/redis_benchmarks_specification/__cli__/args.py
+++ b/redis_benchmarks_specification/__cli__/args.py
@@ -6,7 +6,7 @@
 import datetime
 import os
 from redis_benchmarks_specification.__common__.env import (
-    parse_bool_arg,
+    parse_bool,
     GH_REDIS_SERVER_HOST,
     GH_TOKEN,
     GH_REDIS_SERVER_PORT,
@@ -75,7 +75,7 @@ def spec_cli_args(parser):
     )
     parser.add_argument(
         "--use-git-timestamp",
-        type=parse_bool_arg,
+        type=parse_bool,
         default=True,
         help="Use git timestamp",
     )
@@ -158,7 +158,7 @@ def spec_cli_args(parser):
     )
     parser.add_argument("--id", type=str, default="dockerhub")
     parser.add_argument("--mnt_point", type=str, default="")
-    parser.add_argument("--trigger-unstable-commits", type=parse_bool_arg, default=True)
+    parser.add_argument("--trigger-unstable-commits", type=parse_bool, default=True)
     parser.add_argument(
         "--docker-dont-air-gap",
         default=False,
--- a/redis_benchmarks_specification/__common__/env.py
+++ b/redis_benchmarks_specification/__common__/env.py
@@ -118,6 +118,16 @@ def parse_bool(value, default=False):
             return True
         if normalized in FALSE_STRINGS:
             return False
+        # Warn rather than fall back quietly. The previous bool() read every non-empty string as
+        # True, so an unrecognised truthy-looking value like "2" or "enabled" silently flips to
+        # this default -- and where the flag gates recording benchmark results, that turns a
+        # misconfiguration into missing data rather than an error.
+        logging.warning(
+            "Unrecognised boolean %r; expected one of %s. Using %r.",
+            value,
+            ", ".join(TRUE_STRINGS + tuple(f for f in FALSE_STRINGS if f)),
+            default,
+        )
         return default
     if isinstance(value, (int, float)):
         return bool(value)
--- a/redis_benchmarks_specification/__common__/env.py
+++ b/redis_benchmarks_specification/__common__/env.py
@@ -165,7 +165,14 @@ PULL_REQUEST_TRIGGER_LABEL = os.getenv(
 # store_true flag, so a stray truthy read cannot be overridden on the command line -- setting
 # DATASINK_PUSH_RTS=0 would enable pushing forever. Note the sibling PUSH_RTS in __compare__/args
 # already reads 0 correctly, via int().
-DATASINK_RTS_PUSH = parse_bool(os.getenv("DATASINK_PUSH_RTS"), default=False)
+_DATASINK_PUSH_RTS_RAW = os.getenv("DATASINK_PUSH_RTS")
+# The default deliberately reproduces the old bool() reading -- any non-empty string is True -- so
+# an unrecognised value cannot flip this off. When the flag is False the datasink connection is
+# None and nothing is exported, so a wrong answer in that direction means benchmark results stop
+# being recorded with nothing to indicate why. Only recognised falsy spellings change meaning.
+DATASINK_RTS_PUSH = parse_bool(
+    _DATASINK_PUSH_RTS_RAW, default=bool(_DATASINK_PUSH_RTS_RAW)
+)
 DATASINK_RTS_AUTH = os.getenv("DATASINK_RTS_AUTH", None)
 DATASINK_RTS_USER = os.getenv("DATASINK_RTS_USER", None)
 DATASINK_RTS_HOST = os.getenv("DATASINK_RTS_HOST", "localhost")
--- a/redis_benchmarks_specification/__common__/env.py
+++ b/redis_benchmarks_specification/__common__/env.py
@@ -151,7 +151,11 @@ def parse_bool_arg(value):
 PULL_REQUEST_TRIGGER_LABEL = os.getenv(
     "PULL_REQUEST_TRIGGER_LABEL", "action:run-benchmark"
 )
-DATASINK_RTS_PUSH = bool(os.getenv("DATASINK_PUSH_RTS", False))
+# parse_bool, not bool(): this is the default of a --datasink_push_results_redistimeseries
+# store_true flag, so a stray truthy read cannot be overridden on the command line -- setting
+# DATASINK_PUSH_RTS=0 would enable pushing forever. Note the sibling PUSH_RTS in __compare__/args
+# already reads 0 correctly, via int().
+DATASINK_RTS_PUSH = parse_bool(os.getenv("DATASINK_PUSH_RTS"), default=False)
 DATASINK_RTS_AUTH = os.getenv("DATASINK_RTS_AUTH", None)
 DATASINK_RTS_USER = os.getenv("DATASINK_RTS_USER", None)
 DATASINK_RTS_HOST = os.getenv("DATASINK_RTS_HOST", "localhost")
--- a/redis_benchmarks_specification/__common__/env.py
+++ b/redis_benchmarks_specification/__common__/env.py
@@ -87,6 +87,66 @@ REDIS_BINS_EXPIRE_SECS = int(
     os.getenv("REDIS_BINS_EXPIRE_SECS", "{}".format(24 * 7 * 60 * 60))
 )
 
+TRUE_STRINGS = ("1", "true", "t", "yes", "y", "on")
+FALSE_STRINGS = ("0", "false", "f", "no", "n", "off", "")
+
+
+def parse_bool(value, default=False):
+    """Decode a boolean that has been through a string round-trip.
+
+    Booleans reach us as strings: written to the commits/builds streams with str(), and read from
+    the environment. A bare bool() on that is wrong in the one direction that matters -- every
+    non-empty string is truthy, so "False", "0" and "no" all decode to True and the flag is stuck
+    on. bytes are accepted because stream reads are not always decoded first.
+
+    An unrecognised string returns `default` rather than guessing, since guessing is what bool()
+    already does. Callers that need to distinguish "absent" from "unparseable" should check
+    membership before calling.
+    """
+    if isinstance(value, bool):
+        return value
+    if value is None:
+        return default
+    if isinstance(value, (bytes, bytearray)):
+        try:
+            value = value.decode()
+        except UnicodeDecodeError:
+            return default
+    if isinstance(value, str):
+        normalized = value.strip().lower()
+        if normalized in TRUE_STRINGS:
+            return True
+        if normalized in FALSE_STRINGS:
+            return False
+        return default
+    if isinstance(value, (int, float)):
+        return bool(value)
+    return default
+
+
+def parse_bool_arg(value):
+    """argparse `type=` for a boolean flag.
+
+    Same accepted spellings as parse_bool, but unrecognised input raises so argparse turns it into
+    a usage error instead of silently substituting a default -- on a command line, a typo should be
+    reported, not guessed at. The empty string is rejected for the same reason.
+
+    Accepts exactly the spellings distutils.util.strtobool did, which this replaces: distutils was
+    removed from the standard library in 3.12 and only resolves today via a setuptools shim, so
+    importing it made the CLI fail at import time in any environment without setuptools.
+    """
+    normalized = str(value).strip().lower()
+    if normalized in TRUE_STRINGS:
+        return True
+    if normalized and normalized in FALSE_STRINGS:
+        return False
+    raise ValueError(
+        "invalid boolean {!r}; expected one of {}".format(
+            value, ", ".join(TRUE_STRINGS + tuple(f for f in FALSE_STRINGS if f))
+        )
+    )
+
+
 # environment variables
 PULL_REQUEST_TRIGGER_LABEL = os.getenv(
     "PULL_REQUEST_TRIGGER_LABEL", "action:run-benchmark"
--- a/redis_benchmarks_specification/__common__/env.py
+++ b/redis_benchmarks_specification/__common__/env.py
@@ -128,7 +128,9 @@ def parse_bool(value, default=...):
         try:
             value = value.decode()
         except UnicodeDecodeError:
-            value = None
+            # Left as-is rather than blanked to None: blanking erased the offending bytes from the
+            # message, making an undecodable field indistinguishable from an absent one.
+            pass
     if isinstance(value, str):
         normalized = value.strip().lower()
         if normalized in _TRUE_STRINGS:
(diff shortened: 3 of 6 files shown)


fcostaoliveira:
Pushed `0d98e7e`. Two updates, one of which corrects the reasoning behind the whole boolean investigation.

### The env-var fix can no longer turn pushing off

`parse_bool` returned `False` for anything outside its table, so a deployed `DATASINK_PUSH_RTS=2` or `=enabled` would have flipped from enabled to disabled. When that flag is False the datasink connection is `None` and nothing is exported — so that direction means benchmark results **silently stop being recorded**, which is worse than the bug being fixed.

The default now reproduces the old `bool()` reading, so only recognised falsy spellings change meaning:

```
unset / ""              False -> False
"0" "false" "no" "off"  True  -> False   <- the fix
"1" "true" "yes"        True  -> True
"2" "enabled" "TRUE"    True  -> True    <- unchanged; this was the risk
```

I did this because I **could not verify the fleet's actual setting** — the variable appears in no config-as-code in either repo and the documented deployment passes `--datasink_push_results_redistimeseries` explicitly, but the live supervisor configs aren't readable from where I'm working. Rather than assume, the direction that could lose data is now unreachable. `parse_bool` also logs the value and the accepted spellings when it falls back, so a misconfiguration is visible instead of silent.

### I was wrong about why commit-dating can't be turned off

This investigation started from the belief that the `bool(str)` decode is what makes `--use-git-timestamp false` ineffective. **It isn't**, and I've filed the real cause as #464.

Two independent breaks sit on that value, and the decode is the harmless one:

- `__cli__/cli.py` and `__spec__/cli.py` pass the **literal `True`** to the call that builds the stream payload, and the one place `args.use_git_timestamp` is written targets `reply_fields`, which is not the xadd payload and is reassigned wholesale afterwards. So the flag is discarded before parsing matters.
- The decode bug is real but **inert**: both consumers require `use_git_timestamp is True` *and* `git_timestamp_ms is not None`, and across the three live streams — 22,449 entries, full retention — **zero** entries pair `"False"` with a timestamp. The webhook path that writes `"False"` passes no `gh_token`, so no timestamp is ever set.

There is also a laundering step that makes a partial fix pointless: the builder decodes the field and re-emits `str(value)`, so `"False"` on the commits stream becomes `"True"` on the builds stream regardless.

**This is why the stream-field decode sites are still not in this PR.** Fixing them is a provable no-op on current data, which is a fine reason to do it as hardening but a bad reason to bundle it with a behaviour change. #464 is the one that actually makes the flag work, and it belongs on its own because it moves CLI-triggered datapoints from commit-date to ingest-time for anyone passing `false`.

Related context for whoever picks up the decode sites: `restore_build_artifacts` has **never functioned** — there is no CLI argument for it, the default is hardcoded `True`, the stream field is the only override, and it is only ever written as `"False"` (23 live entries, all with an empty `build_artifacts` list, so the restore it wrongly performs is an empty loop). And `docker_air_gap` is read from the stream but **no writer exists** — 0 of 22,449 entries carry it.

Also filed #465 for nine unguarded `int(os.getenv(...))` calls at import time, same family, crash rather than mis-parse.

101 tests. Reverting the safe default fails 4.

fcostaoliveira:
Pushed `3da5bbe` and `cb613ea`, and rewrote the description. Four reviews, and they were right about more than I expected — including two false claims in my own body and a bug they surfaced *in this PR's own helper*.

### Two claims of mine were wrong

**The `distutils` justification was overstated.** I said the import "survives on an implicit dependency" and that CI isn't using the locked version. Both false: `setuptools` is an **explicit, version-bounded runtime dependency** (`pyproject.toml:38`), and while `tox.yml` does upgrade it unpinned, the **testenv pins `69.5.1` on all six legs**. There's also no scheduled removal of the shim. So this is hygiene — dropping a dependency on a deprecated shim — not the "whole CLI fails to start" hazard I described. Corrected in the body.

**The test count said 80.** It's 144. That's the second body-accuracy hit on this PR and the third across #461/#463; I've stopped treating the body as something to update afterwards.

Also: "accepts exactly the spellings `strtobool` did" is a **superset** — whitespace is tolerated. The suite now pins the tables against `strtobool` itself, so the claim is tested rather than asserted.

### The two-function split was a trap

`parse_bool` was the shorter, first-defined, more discoverable name and the *wrong* one for argparse: as a `type=` callable it swallowed typos, and its own `default=False` overrode argparse's declared `default=True` on the two flags that have one. There's now one function whose safe form is the obvious name — omit `default` and unrecognised input raises; pass one and it warns and falls back.

It raises `argparse.ArgumentTypeError`, not `ValueError`. argparse reproduces the former verbatim and replaces the latter with `invalid parse_bool value: ...`, so the careful list of spellings was being discarded and an internal function name shown instead.

### A regression I'd introduced, and a bug in my own sentinel

`--flag ""` had gone from `False` to `exit 2`. That's what `--verbose "$VERBOSE"` produces with the variable unset, so it now stays `False`. For a stream or environment value, empty means "not supplied" and takes the caller's default — several flags still to be converted default `True` and must not be flipped off by an absent field.

And the review process surfaced a genuine bug in the helper: a module-level `object()` sentinel **does not survive `importlib.reload`**. The signature default is captured at definition time while the identity check reads the module global, so after a reload the strict form silently stopped raising. Not hypothetical — it made the number-rejection tests pass alone and fail in the suite. Now `Ellipsis`, a builtin singleton.

### The guard was the weakest part

The text scan for `type=bool` caught **two of nine** ways to reintroduce the defect. It missed `builtins.bool`, `eval("bool")`, a `functools.partial`, an alias, a tab, a `**kwargs` splat — and a kwarg that `black` reformats across lines, which needs no intent to evade. It also read a relative path, so it passed vacuously from any directory but the repo root.

It now **introspects the built parsers** (`action.type is not bool`), which catches all nine and checks the object argparse will actually call. `--trigger-unstable-commits` also gained behavioural tests; the scan had been its only cover.

The environment tests use subprocesses instead of `importlib.reload` — which fixes the order-dependence *and* reaches what production ships: `__runner__/args.py` and `__self_contained_coordinator__/args.py` take `DATASINK_RTS_PUSH` by value as a `store_true` default, and that default is the entire reason the value matters. The old test asserted a reloaded module constant instead.

### A separate find, filed as #466

Building every parser in the package for the introspection guard turned up something worse than anything in this PR: **the `redis-benchmarks-spec` console script cannot start.** Its entrypoint is `...__spec__.cli:main`, and the package contains a subpackage named `__spec__` which shadows the `ModuleSpec` CPython's own import machinery reads — so importing anything beneath it raises `AttributeError: ... has no attribute '_uninitialized_submodules'`. Pre-existing on `main`, and invisible to CI because nothing in the suite imports it. The same shadowing breaks Flask's `create_app`, which reads `.origin` off that attribute.

That's why `__spec__/args.py` is absent from the introspection guard, with a test asserting the import fails so the dropped coverage gets restored when #466 lands.

### Also

Three other boolean idioms in `env.py` now use the same decoder — `PROFILE` was `bool(int(...))`, which *raises at import time* on `false`, and `BENCHMARK_PR_DIFF_SCOPING` was a hand-rolled copy accepting a different set. Leaving a competing decoder in the same file undercut the point.

Version bumped to `0.3.30`. **Correcting myself again:** I claimed on #461 that bumps land as their own commit and that #459 didn't bump. #459 *did* bump in-PR (0.3.28 → 0.3.29), so in-PR is the convention. I'll note that on #461 too.

144 tests in the file, 172 across the offline set, order-independent in both directions.

fcostaoliveira:
Pushed `5b9d24a`. This round found the most serious defect in the PR so far, and it was mine.

### `PROFILE=2` silently disabled profiling

`PROFILE` was `bool(int(...))`, which treats **every non-zero integer** as enabled. I narrowed it to the literal `"1"`, so `PROFILE=2`, `PROFILE=-1` and `PROFILE=01` flipped from on to **off**. On the profiler runner that means perf and vtune artifacts stop being collected — the box's entire purpose — with nothing to say why.

That is precisely the data-losing direction I had gone out of my way to make unreachable for `DATASINK_PUSH_RTS`, given the opposite policy in the same commit for no stated reason. The default now reproduces the old integer reading, so only the values that previously **crashed at import** change meaning:

```
PROFILE=0 1 2 -1 01     unchanged
PROFILE=false off "" …  ValueError at import  ->  correct bool
```

### Two more of my own inconsistencies

**Undecodable bytes were blanked to `None`**, which erased the offending value from both the warning and the raised message — so a corrupt stream field read identically to an absent one, losing the one diagnostic a reader needs. They now warn with the bytes intact.

Worth recording the wrong turn: I first responded to "warn on every fallback path" by dropping the `None` guard entirely, which made **every unset environment variable log a warning on every import**. `None` means nothing was supplied; that is the normal case and warning about it trains people to ignore the warning that matters. The distinction is now explicit in the code and pinned both ways.

**`BENCHMARK_PR_DIFF_SCOPING`'s `default=False` was asserted in a comment and covered by nothing** — flipping it passed all 164 tests. A `True` there skips no scoping and runs the full suite on every labelled PR. My claim that "each behavioural change was verified by reverting it" was not true for that one. It is now pinned, as its `DATASINK` counterpart already was.

### `__spec__/args.py` — dropped on a false premise, now covered

I had excluded it from the introspection guard claiming it "cannot be imported". That is order-dependent, not absolute: it imports fine once a sibling has populated `sys.modules`, which is exactly why the console script (a cold interpreter) is broken while a test process is not.

But importing it binds the subpackage over the parent's `ModuleSpec`, and Flask reads `.origin` off that to locate an app's root path — so covering it in-process broke `test_app.py` whenever the two files shared a session, in that order. It is covered in a **subprocess** instead, both the flag behaviour and the decoder guard. That also means the introspection guard now spans every parser in the package.

### Docstrings that outlived their claims

Three, all mine: the distutils "hard failure of the entire command" line I had already retracted in the description; a `.decode()` history that described an earlier draft of this PR rather than the base; and a test named "exactly strtobool's" that asserts a superset — now named for that, and asserting the whitespace direction too.

### State

164 tests in the file, **192 across the offline set, identical in both orders** — the order-dependence is gone. All four previously surviving mutants now fail.

Two items from the review I have **not** acted on, with reasons:

- **`--use-git-timestamp ""`** moved from a usage error to `False`, because that flag used `strtobool` while the other eight used `type=bool`. Rather than special-case one flag, I unified on `False` and **corrected the false rationale** that claimed `bool("")` was already `False` there. It is moot in practice — the value is discarded before it reaches the stream (#464) — but it is a real difference from the base, so it is now pinned by a test rather than buried in a comment.
- **`0.3.30` collides with #460**, which also bumps and also edits `env.py` (`+import re` against `+import argparse`). Whichever lands second re-bumps and resolves; flagging rather than guessing the order.

fcostaoliveira:
Pushed `c53d1ab`. Two reviews landed together and both found the same thing, which is the finding that matters most here: **an earlier revision of this branch shrank the guard rather than strengthening it, and the commit message claimed the opposite.**

### I replaced the guard instead of supplementing it

The source scan walked every `.py` in the package. My introspection guard covered two parsers. Injecting `type=bool` into `__runner__/args.py` passed all 164 tests — six of the eight `add_argument` modules had lost all coverage, while `cb613ea` said "building every parser in the package". That claim was false when I wrote it.

All three guards now run, each documented with what only it can see:

| guard | covers |
|---|---|
| introspection | the five parsers buildable in-process |
| subprocess | `__spec__`, which poisons Flask's app-root lookup if imported here |
| AST scan | all eight modules, including `__builder__` and `__api__`, which build inline in `main()` |

Verified by injecting into each of the six previously-uncovered modules in turn.

The scan is now **AST-based**, because the text scan matched its own explanatory prose — it flagged `env.py` at a comment *discussing* `type=bool`. Matching `add_argument(type=...)` in the syntax tree also catches `builtins.bool` and survives reformatting.

### The converter check is a behavioural probe now

Forbidding the name `bool` catches one spelling. A helper returning `bool(x)`, a lambda, a callable class and a shadowed name all reintroduce the defect and all pass `is not bool`. The invariant is that **no converter anywhere may turn the string `"false"` into `True`** — which needs no allowlist to maintain. Worth noting I tried an allowlist of permitted types first and it immediately flagged the legitimate `--from-date` and `--to-date` parsers; the probe has no such upkeep. All four evasions now fail.

Also fixed: the flag check required a converter on `store_true` actions, which consume no value and legitimately have none.

### The default direction had no coverage — on this branch's own safety argument

Flipping the six compare flags to `default=True`, or an unset `DATASINK_PUSH_RTS` or `PROFILE` to `True`, **passed the entire suite**. An unset `DATASINK_PUSH_RTS` reading `True` would start pushing to the datasink from a process that never asked to — the mirror image of the failure this branch exists to prevent, and the direction I claimed to have made unreachable. Every environment case set a value; none checked the absent case, which is what almost every process actually runs with. All three now fail. `_env_probe` also clears every variable it can be asked about, so "unset" is genuinely unset regardless of the developer's shell.

### Also

`PUSH_RTS` in `__compare__/args.py` was `bool(int(os.getenv(...)))` — the same crash-on-a-word defect I fixed for `PROFILE`, in a file this branch already edits and to which it had already added `parse_bool` to the imports. `PUSH_RTS=false` raised `ValueError` at import, aborting every compare command. Fixed with the same policy: no non-zero value flips to disabled.

The distutils check reads `sys.modules` rather than source text — a source check passes on `importlib.import_module("distutils.util")` and fails on the word appearing in a comment.

**Version moved to `0.3.31`.** #460 claimed `0.3.30` an hour earlier, and the two produce **no conflict in `pyproject.toml`** because both add the identical line — so the second to merge would have shipped silently under an already-published number. #460 holds the prior claim, so this one yields.

### Not done, with reasons

- **Trimming the ~39 redundant parametrize cases.** Fair, and I've stopped quoting the total as a quality signal — it is a parametrize artifact, not 181 ideas. But deleting coverage in the same round that restored coverage is the wrong order of operations; worth a follow-up once this is settled.
- **Converting 16 of the 21 subprocess cases** to a fresh-module load via `spec_from_file_location`. The measurement is convincing (82% of runtime in 15% of tests, ~10x faster with an identical kill set) and I'd take it — but it changes how every environment test reaches its subject, and doing that in the same push as the coverage restoration would make both hard to review. Also a follow-up.
- **The `__spec__` tripwire.** Both reviews want it `xfail(strict=True)` or moved so it doesn't red a stranger's PR when #466 lands. I agree it shouldn't fail someone else's work from a file called `test_parse_bool.py`; I'd rather settle where it lives than move it twice.

181 tests in the file, 209 across the offline set, identical in both orders.

PR #470, opened 2026-08-03
fix: honour the per-run priority limits on the builds stream, and make the stream read total

### LINKED ISSUE #448: Per-run priority limits are silently ignored: builds stream writes tests_priority_* but the coordinator reads priority_*
## Summary

`--tests-priority-upper-limit` / `--tests-priority-lower-limit` have no effect on any run. The producer and the consumer disagree on the stream field name, and because the consumer uses an `if key in payload` membership test, the mismatch is not an error — it silently falls back to the coordinator's own default.

## Evidence

The value is written with a `tests_` prefix and read without one:

| Step | Location | Field name |
|---|---|---|
| CLI → commits stream | `redis_benchmarks_specification/__cli__/cli.py:605-606` | `tests_priority_upper_limit` |
| builder reads commits stream | `__builder__/builder.py:421-428` | `tests_priority_upper_limit` ✅ |
| builder → builds stream | `__builder__/builder.py:968-969` | `tests_priority_upper_limit` |
| coordinator reads builds stream | `__self_contained_coordinator__/self_contained_coordinator.py:1285` | **`priority_upper_limit`** ❌ |

```python
# __builder__/builder.py:968
"tests_priority_upper_limit": tests_priority_upper_limit,

# __self_contained_coordinator__/self_contained_coordinator.py:1285
if b"priority_upper_limit" in testDetails:
```

Consequence: the branch is never taken, `priority_upper_limit` keeps the value derived from the coordinator's own CLI args, and the log line

```
detected a priority_upper_limit definition on the streamdata ... will replace the default upper limit of ...
```

has never been emitted. The same applies to the lower limit.

## Impact

Any run that relies on the per-run priority window to scope which tests execute has silently run with the coordinator's default window instead. Runs that appeared capped were not capped.

## Why it was not caught

The two components communicate through bare string literals on both sides with nothing linking them, and `if b"name" in payload` turns a name mismatch into a silent default rather than a failure. There is currently no test asserting that a field a consumer reads is a field a producer writes.

## Proposed fix

1. Declare the stream field contract in one place, and read through a helper that resolves documented aliases and reports which spelling it matched (so an alias hit is visible in logs rather than silently equivalent).
2. Treat `tests_priority_*` as canonical, since that is what every producer has always written, and accept the un-prefixed spelling as an alias so entries already in flight — and coordinators still running an older pinned release — are honoured. No flag day required.
3. Add cheap tests (no Redis, no Docker) that assert: every consumed field is declared; consumed-but-never-produced fields match a recorded set exactly; and callers pass canonical names rather than aliases.

Worth noting for reviewers: **this is a behaviour change.** Once fixed, the cap starts taking effect, so results gathered previously with `--tests-priority-*` are not necessarily comparable with results gathered after.

A branch implementing the above is ready and I can open it as a PR against this issue.

## Related findings while investigating

Two other stream fields are read but never written by any in-repo producer, so those settings are likewise unreachable. Filing separately to keep this issue focused.

fcostaoliveira:
Closes #448.

The trigger CLI, the builder and the coordinator talk only through Redis Streams, using bare string literals as field names on both sides. A rename or typo there does not fail — the consumer's `if b"name" in testDetails` check is simply `False` and it silently keeps its own default.

That is not hypothetical. The builder writes `tests_priority_upper_limit`; the coordinator read `priority_upper_limit`. So `--tests-priority-upper-limit` and `--tests-priority-lower-limit` **had no effect on any run**, and the `detected a priority_upper_limit definition on the streamdata` log line has never once fired. Historical results gathered with those flags were uncapped, so they are not necessarily comparable with post-fix runs.

## The fix

`__common__/stream_contract.py` declares the commits and builds field sets in one place, with the documented aliases, and `read_stream_field()` replaces the hand-rolled `if b"x" in d: int(d[b"x"].decode())` at every call site — which is where the drift crept in.

It returns `(value, matched_name)`. `matched_name` is what lets a caller distinguish "the producer said `0`" from "the producer said nothing", and it makes an alias hit visible in the logs instead of silently equivalent. The coordinator now reads the priority limits through it.

A per-PR workflow guards the contract: it asserts every `read_stream_field()` call in the package uses a declared canonical name, so an alias or a typo fails in seconds rather than becoming a setting that quietly stopped working. Offline — no Redis, no Docker, no fleet.

## What fuzzing then found

The unit tests pin known cases. A property test over generated payloads found one nobody had written a case for:

```
payload = {b'arch': b'\x80'}
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0
```

`read_stream_field` is documented as total on the no-cast path, and it is read inside consumer loops that **ACK unconditionally** — so raising there drops the work silently and never retries it. Strictly worse than a value the caller can see.

Undecodable bytes are now decoded lossily and **logged**, not treated as absent:

- *Not absent*, because a corrupt value is not a missing one. Folding them together would erase the distinction `matched_name` exists to preserve — a caller could no longer tell "the producer wrote garbage" from "the producer wrote nothing".
- *Logged*, because that is what keeps a lossy decode from being a silent corruption. Without it the caller sees a mangled string and nothing anywhere says why.

`bytearray` is accepted alongside `bytes`, since redis-py is not the only producer of these mappings.

## The nightly fuzz job could never have run

Both defects measured, not inferred:

**`--timeout=1800` with only `pytest hypothesis` installed.** That option comes from `pytest-timeout`, so every one of the eight shards exited **4** with `unrecognized arguments: --timeout=1800` before evaluating a single example. Now installed.

**`PYTHONHASHSEED` does not seed Hypothesis.** It was set per shard to "explore different inputs instead of running the same examples eight times". Verified: the *same* `PYTHONHASHSEED` twice produces *different* example sequences, so the shards diverged only by run-to-run randomness — and nothing found at night could be replayed. `--hypothesis-seed=<shard>` is reproducible (verified identical across runs), so the shards now genuinely differ *and* the shard number alone reproduces a finding.

The nightly is sharded eight ways with a per-shard Hypothesis example database, cached with `restore-keys`, so each night starts from the previously-found interesting inputs and coverage compounds rather than restarting from scratch. `fail-fast: false` so one shard failing cannot hide the others, and falsifying examples upload as artifacts.

## Tests

16 in the contract suite — 10 unit cases plus 6 properties. Offline, seconds on a PR, dialled up nightly via `HYPOTHESIS_MAX_EXAMPLES`.

The properties are chosen for the bug class rather than for coverage: that the no-cast read never raises on arbitrary producer input; that a reported `matched_name` is really a key in the payload (so logs cannot claim something the wire did not carry); that the canonical spelling beats an alias *deterministically*, since ambiguous precedence would make behaviour depend on dict ordering; that absence reports as absence rather than as a value; and that random keys never accidentally satisfy a declared lookup.

Each change verified by reverting it: dropping the decode guard fails 2, dropping the warning fails 1.

---

> [!NOTE]
> **Medium Risk**
> Changes which benchmark tests run when priority flags are set (behavior fix, but prior results may not be comparable). Touches core stream orchestration paths; mitigated by extensive contract and property tests on every PR.
> 
> **Overview**
> **Fixes a long-standing silent bug** where the coordinator read `priority_upper_limit` / `priority_lower_limit` while producers wrote `tests_priority_upper_limit` / `tests_priority_lower_limit`, so CLI flags like `--tests-priority-upper-limit` never affected runs.
> 
> Introduces **`__common__/stream_contract.py`** as the declared commits/builds field contract, with **`read_stream_field()`** (aliases, safe `cast=int`, lossy decode + warnings instead of raising in consumer loops). The **coordinator** and **builder** use it for priority limits; the builder **omits** priority fields on the builds stream when unset (so it does not override coordinator defaults) and **forwards** them only when present. The **dockerhub CLI** now passes `tests_groups_regexp` and priority limits instead of hardcoded values.
> 
> Adds a **Stream contract** GitHub Actions job running unit + Hypothesis fuzz tests (fails on skipped tests). **`__init__.py`** version fallback is broadened so missing `pkg_resources` does not break test collection.
> 
> Also adds several **memtier benchmark YAML** specs (hash `hash-min-template-entries=1` template variants, including feature-store playbooks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 643e4c015ccd4a5f2ee7e4076db1d740cb5a952a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Pushed `c50dbe8` and `643e4c0`. Two reviews, and between them they showed the previous state of this PR **traded one silently-ignored setting for another and added a run-dropping path** — plus my own workflow "fix" made that workflow worse. Ten findings acted on.

### The patch reintroduced its own failure mode, pointing the other way

The builder wrote the priority fields **unconditionally**, so once the coordinator honoured them there was no way to distinguish "the operator asked for 100000" from "argparse defaulted to 100000". A per-runner cap set with `--tests-priority-upper-limit` was therefore overwritten on **every run**, with an INFO line claiming a definition had been detected on the stream. Measured: base gives `[0, 3]` (operator honoured), the PR gave `[0, 100000]` (operator silently overridden).

Absence now stays absence across the hop — the builder reads with no default and omits the field, following the pattern the payload builder already used for its other optional fields.

### The totality fix achieved nothing at either real caller

Both live callers pass `cast=int`. The lossy decode ran, logged *"decoding lossily"*, and then `int()` raised anyway — so the warning was actively misleading. And it was **newly reachable**: on main the coordinator's key was never present, so `int()` never ran; after the fix it is present on every entry, at a site whose handler logs CRITICAL and whose caller then **ACKs unconditionally**. A malformed priority field discarded a whole benchmark run and never retried it.

A failing cast is now reported as absence, with a warning, degrading to the caller's own default exactly as a missing field does. The docstring says so — the previous comment cited a promise the docstring never made.

### The flags still didn't work on the dockerhub producer

`__cli__/cli.py` passed the literals `".*"`, `0`, `10000` into slots 19–21 of a **31-argument positional call**, discarding `args.tests_groups_regexp` and both priority limits. `spec_cli_args` is shared by every subcommand, so the flags were accepted there and dropped — on the producer the dockerhub and GA-tag baseline triggers actually use. The review's slot mapping was exact; I confirmed it by AST.

### The job I added was red, and I should have checked before dispatching reviews

Importing the package runs `__init__.py`, whose version fallback names `pkg_resources.DistributionNotFound` in an `except` clause — dereferencing the module whose import just failed. So the intended `99.99.99` fallback raised `NameError`, and any job that doesn't install the package dies at collection. setuptools 83 no longer ships `pkg_resources`, so that's the common path now. Fixed at the source.

### I deleted the nightly rather than repairing it

My own "fix" to it was worse than the bug: `--hypothesis-seed` sets `database=None`, so the example database the cache and upload-on-failure steps exist to preserve was **empty by construction**. "Coverage compounds" and "falsifying examples upload" were both false, and a fixed seed with no database replays one identical sequence every night, forever.

And the measurement settles it: branch coverage saturates at **200** examples and the mutation score is **identical at 200 and 20000**. Eight shards × 20000 examples ≈ 292 runner-hours a year for nothing detectable. The properties now run at 500 on every PR — where a regression is cheap to fix, since a nightly cannot block the PR that introduces one.

The job also **fails if any test was skipped**: the properties `importorskip` on hypothesis, and in a multi-file session a skip still exits 0, so the build would have gone green with every property absent.

### The properties killed 4 of 9 mutations. Now 7 of 7.

Two deleted as worthless:

- `test_unrelated_payloads_never_produce_a_match` asserted random binary keys don't equal a field name — about a 2⁻³² chance, so all 8200 assertions were vacuous. It was also the most expensive property, roughly a quarter of the runtime.
- `test_int_round_trip` built a payload containing exactly the canonical key and read it back. Restated the implementation; killed nothing.

Five added, chosen from what survived — most importantly **a key on the wire is always reported as matched**. Without that converse direction the entire lookup is deletable: returning `(default, None)` unconditionally satisfied every original property. Also: an alias-only payload still matches and reports the alias (deleting the alias lookup — the module's stated compatibility contract — previously passed at 20000 examples); a decodable value is returned byte for byte; an empty value is a value, not an absence; and a bytes key takes precedence over a str key.

Also fixed a dead generator arm: the `surrogatepass` encode was gated on `.isprintable()`, never true for a lone surrogate, so every string needing the handler was mapped to `b""` and all undecodable coverage came from `st.binary` alone.

### Corrections to the description

- **"replaces the hand-rolled read at every call site"** was wrong by about 26×. Two sites were converted; 53 `b"..." in testDetails` reads remain, including five `int(...decode())` — the exact idiom named as the root cause. The body now says two.
- **The nightly-workflow findings were intra-branch churn**, not a live outage. The workflow was introduced by the first commit on this branch and fixed by the second; presenting that as a discovery was misleading, and the whole file is gone now anyway.
- **Five fields were produced but undeclared** — `ref`, `gh_org`, `gh_repo`, `os`, `archived_zip`. Now declared. The module claimed "skipping step 1 makes the contract test fail"; producers are **not** held to the set, and the docstring now says which direction is enforced and which is not.

An undeclared or alias name now raises inside `read_stream_field`. The source-regex guard it replaces was defeated five ways — single-quoted literal, keyword argument, module constant, f-string, subscripted first argument — each reinstating the original inert read with CI green.

24 tests. Each fix verified by reverting it.

PR #483, opened 2026-08-12
fix: port init_lua bulk population to preload_tool for multi-shard portability

### LINKED ISSUE #482: [BUG]: init_lua-populated suites silently produce a mispopulated dataset on multi-shard deployments
**Describe the bug**

Nine test-suites populate their dataset with `dbconfig.init_lua` — a single `EVAL` invoked with `numkeys=0` whose body loops over and writes N distinct keys. This makes them unusable on any deployment where the keyspace is distributed across more than one shard, and — worse — the failure is **silent**.

A script invoked with `numkeys=0` carries no key hints, so it executes in the context of a single shard. Its `redis.call('SET'/'ZADD'/'JSON.SET', ...)` writes therefore cannot populate keys that belong to other shards. The benchmark phase then routes its reads by key and finds nothing. No error is raised, the run completes, and the throughput number that comes out is mostly the cost of returning misses.

Observed on a multi-shard deployment:

| Suite | miss rate during the benchmark phase |
|---|--:|
| `memtier_benchmark-playbook-session-storage-100k-sessions` | **~94%** |
| `memtier_benchmark-playbook-session-caching-string-100k-sessions` | **~33%** |

The two differ only because of their read/write mix: the session-caching suite sends ~10% `SET`s, which land correctly through the normal data path and progressively repopulate part of the keyspace during the run. The read-heavy session-storage suite has almost no such self-healing and stays broken for the whole measurement.

Both suites still reported a successful run with a valid-looking ops/sec figure. A consumer that does not separately inspect `Misses/sec` will treat those numbers as real.

**The same repo already contains the portable pattern**

This is not a limitation of the workloads — it is an inconsistency in how sibling suites in the *same* playbook families are written:

| Family | Suite | Population | Portable? |
|---|---|---|---|
| session-caching | `...-hash-100k-sessions` | `preload_tool` | ✅ |
| session-caching | `...-string-100k-sessions` | `init_lua` (100,000 keys) | ❌ |
| session-caching | `...-json-100k-sessions` | `init_lua` (100,000 keys) | ❌ |
| leaderboard | `...-top-100` | `preload_tool` | ✅ |
| leaderboard | `...-top-10` / `...-top-1000` | `init_lua` (10 / 1,000 keys) | ❌ |
| session-storage | `...-100k-sessions` / `...-1k-sessions` | `init_lua` (100,000 / 1,000 keys) | ❌ |

`session-caching-hash-100k-sessions` and `leaderboard-top-100` do exactly the same job with `preload_tool` and work correctly regardless of topology. They are ready-made templates for the others.

**Full list of affected suites**

Population done entirely in `init_lua` (no `preload_tool`) — 9 suites:

- `memtier_benchmark-playbook-session-caching-string-100k-sessions`
- `memtier_benchmark-playbook-session-caching-json-100k-sessions`
- `memtier_benchmark-playbook-session-caching-hash-100k-sessions-with-hfe-expired-fields`
- `memtier_benchmark-playbook-session-caching-hash-100k-sessions-with-hfe-expired-fields-hkeys`
- `memtier_benchmark-playbook-session-storage-100k-sessions`
- `memtier_benchmark-playbook-session-storage-1k-sessions`
- `memtier_benchmark-playbook-leaderboard-top-10`
- `memtier_benchmark-playbook-leaderboard-top-1000`
- `memtier_benchmark-1Kkeys-hash-listpack-500-fields-update-20-fields-with-1B-to-64B-values`

A further 5 suites use `init_lua` *alongside* `preload_tool` (the `1Mkeys-info-everything-*-cmdstats` set). Those look unaffected — their Lua appears to be small setup rather than bulk key population — but they are worth a glance while this is being fixed.

**To Reproduce**

1. Run `memtier_benchmark-playbook-session-storage-100k-sessions` against a deployment whose keyspace spans more than one shard.
2. Let `dbconfig.init_lua` perform the population.
3. Run the benchmark phase and inspect `ALL STATS/Totals` in the memtier JSON.
4. `Misses/sec` dominates `Hits/sec`; the run reports success regardless.

**Expected behavior**

Either of:

1. **Preferred — make the population topology-portable.** Port the `init_lua` bulk population to `preload_tool`, following `session-caching-hash-100k-sessions` and `leaderboard-top-100`. This keeps the suites usable everywhere and removes the failure mode entirely.
2. **At minimum — make the constraint explicit and detectable.** These suites already declare `redis-topologies: [oss-standalone]`, but nothing records *why*, and nothing prevents a silent bad run if they are used elsewhere. A comment in each YAML plus a `dbconfig.check.keyspacelen` assertion after population would turn a silent mispopulation into an immediate, obvious failure.

Option 2 is worth doing even if option 1 lands, since `keyspacelen` guards against any future population regression. Note that `session-caching-hash-100k-sessions` already sets `dbconfig.check.keyspacelen: 100000`, while the `init_lua` suites listed above do not — so the suites most exposed to this are precisely the ones without the check.

**Environment**

- Not environment-specific: the behaviour follows from `EVAL` with `numkeys=0` plus a distributed keyspace.
- Suite definitions as of `redis_benchmarks_specification/test-suites/` on `main`.

**Additional context**

Happy to open a PR converting the `session-caching-string`, `session-storage` and `leaderboard` suites to `preload_tool` and adding `keyspacelen` checks, if the maintainers agree with the direction.

fcostaoliveira:
Fixes #482.

## Problem

9 test-suites populated their dataset via `dbconfig.init_lua` — a single `EVAL <script> 0`. A script invoked with `numkeys=0` carries no key hints, so on a multi-shard deployment it executes entirely on one shard and never populates keys that hash to other shards. The benchmark still reports a successful run, with an ops/sec figure that's actually dominated by cache misses on the un-populated majority of the keyspace.

## Fix

Ports population to `memtier_benchmark`'s own `preload_tool` — a real client that routes every write by its actual key via `--command-key-pattern="P"`, so it works correctly regardless of shard count. This follows the pattern already used by `session-caching-hash-100k-sessions` and `leaderboard-top-100`, both cited in the issue as working references.

Every suite's `dbconfig.check.keyspacelen` is set from an actually-measured `DBSIZE` after preload, not assumed — several existing off-by-one conventions in this repo (`--key-maximum` inclusive/exclusive depending on flag combination) are easy to get wrong by eyeballing.

Suites converted (the full list from the issue):
- `session-caching-string-100k-sessions`
- `session-caching-json-100k-sessions`
- `session-caching-hash-100k-sessions-with-hfe-expired-fields` (+ its `-hkeys` sibling)
- `session-storage-100k-sessions` / `session-storage-1k-sessions`
- `leaderboard-top-10` / `leaderboard-top-1000`
- `1Kkeys-hash-listpack-500-fields-update-20-fields-with-1B-to-64B-values`

## Also included: two independently-discovered bugs, same failure class

Found while using `leaderboard-top-100` as the reference template — bundled here rather than split out since they're one-line/one-flag fixes in files this PR already touches for the same reason (silent preload/measurement key mismatch):

- **`leaderboard-top-100` and `leaderboard-top-1000`** preloaded without `--key-prefix ""`, so memtier defaulted to writing `memtier-<id>` keys while `clientconfig` reads bare `<id>` — a 100%-miss bug independent of the multi-shard issue above, on the file that's supposed to be this whole PR's proof that `preload_tool` works.
- **`leaderboard-top-10`/`leaderboard-top-1000`** issued 200x more preload requests than needed: `-n 10000` at `-c 50 -t 4` (200 connections) walks the full range on every connection rather than each connection's `P`-partitioned share. Harmless on `top-10`, but `top-1000`'s ~1000-pair-per-command payload made this slow enough to risk the fleet's fixed-work container timeout. Switched to the exact per-connection request count (`key-range ÷ connections`).

## Validation

Every preload was run locally against real Redis + the `:edge` memtier image and checked for exact `DBSIZE`/`OBJECT ENCODING`/field-content match against what each suite's (untouched) `clientconfig` reads — not just eyeballed. The two HFE suites' HSET-then-HPEXPIRE ordering was checked via `HTTL` across 50+ scattered keys (not `HLEN`, which lags behind Redis's active-expire sweep by several seconds and gives false positives). `clientconfig`, `tested-commands`, `tested-groups`, `redis-topologies`, `build-variants`, and `priority` are unchanged in every file.

---

> [!NOTE]
> **Medium Risk**
> Changes how benchmark datasets are populated and how client load is applied, which can alter measured throughput and invalidate historical comparisons if preload or keyspace checks are wrong.
> 
> **Overview**
> Replaces `init_lua` dataset seeding with `memtier_benchmark` `preload_tool` so keys are routed by actual key hash and populate correctly on multi-shard deployments.
> 
> Converts the large-hash update suite and `leaderboard-top-10` from Lua `EVAL` preload to partitioned memtier writes (`--command-key-pattern="P"`), and adds an explicit `keyspacelen` check where missing.
> 
> Also fixes silent preload/client key mismatches on the leaderboard suites: adds `--key-prefix ""` on `leaderboard-top-100` preload, and retunes client load from `-c 50 -t 4` / 4 CPUs to `-c 25 -t 8` / 8 CPUs across the touched suites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49018a7c31a47d3fed43bf1f0182542a27a7db03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Pushed a follow-up commit: bumped `clientconfig` from the 4-threads/50-conns-per-thread default to 8 threads/25 conns-per-thread (same 200 total connections) across all 10 suites, and `resources.requests.cpus` from `4` to `8` to match.

Measured directly on the CI fleet's `oss-standalone` topology (1 server core cap), the client wasn't the bottleneck at 4 threads — throughput was identical at 4 vs 16 client cores while the server sat at ~100% of its single core. But these suites are meant to be usable beyond that topology too, and 4 client threads can undersupply concurrent per-shard connections once the server side can actually parallelize across more shards. Spreading the same total concurrency across more threads costs nothing on the CI topology and removes that ceiling elsewhere.

PR #484, opened 2026-08-12
test-suites: add SORT / SORT_RO benchmark coverage (#469)

### LINKED ISSUE #469: Add benchmark coverage for SORT (SORT / SORT_RO, including BY and GET patterns)
We have 432 specs in `test-suites/` and none of them run `SORT` or `SORT_RO`. Every spec that mentions "sort" is a sorted-set (`Z*`) test — grepping the suite for a memtier command containing `SORT` returns nothing. So one of the more expensive read paths in the server is completely unmeasured.

Why this one is worth covering:

- `SORT` is O(N+M*log(M)). It materializes the whole collection into a vector, sorts it, and replies, so cost scales with the element count of a single key rather than with the number of keys touched. Nothing else in the suite has that shape.
- With `BY <pattern>`, `sortCommandGeneric()` calls `lookupKeyByPattern()` once per element — one command becomes N keyspace lookups plus N object allocations. Every `GET <pattern>` adds another N. That per-element lookup driven from inside a single command is the part no other spec exercises.
- `BY hash->field` / `GET hash->field` adds a hash field read per element on top, through `hashTypeGetValueObject()`.
- `SORT` on a listpack-encoded sorted set converts it to a skiplist in place, so the first call pays a full re-encode.
- `ALPHA` swaps the numeric compare for `strcoll()`, which is a different cost profile and locale-dependent.

We do have plenty of `list`, `set`, and `sorted-set` specs, so the *types* SORT reads are well covered — it is only the command that never runs.

Suggested additions, smallest useful set first:

1. `SORT` on a 1K-element numeric list, no options — baseline compare and allocation cost.
2. `SORT ... ALPHA` on a 1K-element list of short strings — the `strcoll()` compare path.
3. `SORT ... BY weight_* GET # GET data_*` over a 1K-element list — the per-element `lookupKeyByPattern()` path, which is the main thing missing today.
4. `SORT ... LIMIT 0 10` on a large list — guards the common "sort a big collection, return one page" shape.
5. `SORT_RO` mirroring (1), so the read-only variant stays covered as the two diverge.

Element counts are the open question — 1K keeps the per-op latency in a range memtier can drive at reasonable throughput, but if we want the sort itself to dominate rather than the reply, a second tier at 10K would show the scaling. Happy to take this on if nobody else has it queued.

fcostaoliveira:
Closes #469.

Adds 4 specs that give the suite its first coverage of the `SORT` / `SORT_RO`
commands. Until now the entire suite had zero specs invoking `SORT` — every
existing \"sort\" match is a sorted-set (`Z*`) benchmark — so the whole-list
vector build path, the ALPHA `strcoll()` compare, and (most importantly) the
per-element `lookupKeyByPattern()` path driven from BY/GET were completely
unmeasured. This directly showed up when someone asked to benchmark a
`SORT`-affecting change and there was nothing scoped to run for it.

### Specs

All four use a single 1K-element LIST key, `oss-standalone`, pipeline 10 — the
smallest useful set called out at the bottom of #469. The names, dataset
shapes and rationale line up 1:1 with items (1), (2), (3), and (5) of the
issue.

| File | Command | Cost profile isolated |
|---|---|---|
| `memtier_benchmark-1key-list-1K-elements-sort-numeric.yml` | `SORT sort:list:1K:num` | Baseline whole-list vector build + numeric compare + full reply |
| `memtier_benchmark-1key-list-1K-elements-sort-alpha.yml` | `SORT sort:list:1K:alpha ALPHA` | `compareStringObjectsWithOrder()` / `strcoll()` per-element string compare |
| `memtier_benchmark-1key-list-1K-elements-sort-by-get-external.yml` | `SORT sort:list:1K:byget BY weight_* GET # GET data_*` | Per-element `lookupKeyByPattern()` — 1000 BY lookups + 1000 GET lookups per single SORT call. Preloads 2000 companion `weight_<i>` / `data_<i>` string keys via an EVAL, so keyspacelen is 2001. |
| `memtier_benchmark-1key-list-1K-elements-sort_ro-numeric.yml` | `SORT_RO sort_ro:list:1K:num` | Mirrors the numeric baseline for the read-only entry point so future divergence between `SORT` and `SORT_RO` is measurable. |

### Group

`SORT` is classified as `generic` in Redis's own command metadata
(`commands.json: SORT.group == 'generic'`), so all four specs use
`tested-groups: [generic]`.

### Element count

1K matches the \"smallest useful set first\" recommendation in the issue —
per-op latency stays in memtier's driveable range and the sort itself is the
dominant cost rather than the reply write-out. A 10K second tier (as also
suggested in #469) is a natural follow-up but is intentionally left out here
to keep this PR reviewable.

### Preload notes

- The numeric variants use `preload_tool` with `RPUSH __key__` iterating
  1..1000, matching the pattern already used by
  `memtier_benchmark-1key-list-10K-elements-lindex-integer.yml`.
- The ALPHA and BY/GET variants use a single `EVAL` (same shape as the
  existing streams specs) because they need deterministic non-numeric string
  payloads and companion keys — `oss-standalone` only, so cross-slot concerns
  from PR #483 don't apply.

### Triggering only these

Once merged, scoping a benchmark run to just this coverage is:

```
--tests-regexp 'list-1K-elements-sort'
```

---

> [!NOTE]
> **Low Risk**
> Benchmark YAML only; no runtime or Redis server code changes.
> 
> **Overview**
> Adds **four new memtier benchmark specs** so the suite can measure **`SORT`** and **`SORT_RO`** for the first time (closes #469). Each run uses one **1K-element LIST**, **`oss-standalone`**, and **pipeline 10**.
> 
> **`SORT` numeric** and **`SORT_RO` numeric** preload via memtier `RPUSH`, set **`list-max-listpack-size: 128`** for multi-node quicklist, and drive default numeric sort vs read-only entry point.
> 
> **`SORT ... ALPHA`** preloads deterministic alphabetic strings with an **EVAL** and the same listpack tuning to stress string/`strcoll()` compare.
> 
> **`SORT ... BY weight_* GET # GET data_*`** preloads the list plus **2000 companion keys** (`keyspacelen: 2001`) via **EVAL** to exercise per-element **`lookupKeyByPattern()`** on BY/GET.
> 
> All specs tag **`tested-groups: generic`** and **`tested-commands`** as `sort` or `sort_ro`. Runs can be scoped with `--tests-regexp 'list-1K-elements-sort'`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23088561d0d45a98657b2fa087dc667037e08060. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Ran this through a 7-lens adversarial review before merge. Three real bugs came out of it, all fixed in the latest commit:

- **`sort-numeric` / `sort_ro-numeric`**: preload undershot by one element (999 not 1000 — `--key-maximum` off-by-one), and the encoding claim was wrong. 1000 short numeric elements stay `listpack`-encoded under the default `list-max-listpack-size` (an 8KB size cap they never approach) — not the `quicklist` the description claimed. Fixed the off-by-one and added an explicit `list-max-listpack-size: 128` override to force genuine multi-node `quicklist` storage, matching the documented intent.
- **`sort-alpha`**: the Lua preload's `string.sub(chars, math.random(1,26), math.random(1,26))` drew two independent random indices, producing variable-length (0-105 byte, including empty-string) substrings instead of the claimed fixed 8-byte lowercase strings. Fixed to draw one random index per character. Verified empirically — all 1000 elements are now exactly 8 bytes.
- All 4 files renamed with a `-pipeline-10` suffix to match this repo's otherwise-universal naming convention for pipelined specs (every existing `1key-list-*-pipeline-N` spec follows it; these four didn't). Descriptions now state encoding directly rather than only in `dataset_description`.

Two things the review also raised that I'm **not** changing, flagging instead:

- **Item (4) from #469** (`SORT ... LIMIT 0 10` on a large list) isn't covered by this PR — unlike the 10K-tier omission (called out explicitly in the description above), this one wasn't disclosed. It's a real gap worth its own follow-up rather than squeezing into this PR.
- Per this repo's usual pattern (most recently the v0.3.34 release), merging these specs doesn't make them live on the fleet — that needs a version bump + reinstall (or `scp` + `--target-platform`) as a separate step.

Every fix above was validated against a real `redis-server` + `memtier_benchmark:edge`, not just read.

Issue #487, opened 2026-08-12
1key-set-10-elements-smembers-pipeline-10 runs near load-generator saturation (memtier at 81% of its threads)


fcostaoliveira:
**Describe the bug**

`memtier_benchmark-1key-set-10-elements-smembers-pipeline-10` runs close to **load-generator saturation**, so its reported ops/sec is partly a property of memtier rather than of the server.

Measured on a multi-core host with the load generator pinned to cores separate from the server:

| | |
|---|---|
| memtier CPU | **3.22 of its 4 threads — 81%** |
| server-side CPU | ~37% idle capacity remaining |

This suite is the highest-throughput one in the set we exercised (>1.2M ops/sec), which is exactly why the client runs out of room first: at `--pipeline=10` on a small 10-element SET, per-operation client cost is low but operation *rate* is very high, and memtier's 4 default threads become the limit.

**Honest caveat, unlike the two sibling reports**

For the `session-caching-hash` and `feature-store` suites I filed alongside this one, there is a clear in-repo inconsistency to point at — their own family members use `-t 8` / `-t 10`. **Here there is no such precedent**: this suite and the related `playbook-realtime-analytics-membership` both use memtier's defaults, and no comparable set suite requests a higher `-t`. The argument here rests only on the measurement.

81% is also less clear-cut than the 91% of the other two. It may be judged acceptable; it is reported for completeness so the maintainers can decide, rather than because the case is equally strong.

**Expected behavior**

Either raise `-t` (with `resources.requests.cpus` matched, as every suite in the repo does) so the client keeps clear headroom over the server, or document that this suite is intentionally client-limited at high op rates so consumers do not read its ops/sec as a server-capacity figure.

**To Reproduce**

1. Run the suite against a server with spare capacity, memtier co-located but on separate cores.
2. Sample memtier's own CPU (profiler, or `pidstat -p $(pgrep memtier_benchmark)`).
3. memtier sits at ~81% of `-t` threads while the server has idle capacity.

**Additional context**

A general observation from the three reports together: suites that omit `-t` inherit memtier's default of 4, and whether that is enough depends entirely on the workload's op rate. A cheap guard would be for the runner to warn when memtier's own CPU exceeds some fraction of its thread budget during a run — that turns "this number is really a client measurement" from something you have to profile for into something the run tells you.

Issue #486, opened 2026-08-12
feature-store serving suite is load-generator bound (memtier at 91% of its threads); ingest sibling uses -t 10, serving uses -t 4


fcostaoliveira:
**Describe the bug**

`memtier_benchmark-playbook-feature-store-hash-10M-entities-50-features` (the serving/read suite) is **load-generator bound**: memtier is the bottleneck, not the server under test.

Measured on a multi-core host with the load generator pinned to cores separate from the server:

| | |
|---|---|
| memtier CPU | **3.65 of its 4 threads — 91%** |
| server-side CPU | ~33% idle capacity remaining |

At 91% of its own thread budget the client cannot offer more load, so the reported ops/sec reflects memtier's capacity rather than the server's.

**Inconsistent with the ingest suite it pairs with**

| Suite | `-t` | `-c` | `--pipeline` | `resources.requests.cpus` |
|---|--:|--:|--:|--:|
| `...-feature-store-hash-ingest-10M-entities-50-features` | **10** | 10 | 50 | 10 |
| `...-feature-store-hash-10M-entities-50-features` (serving) | 4 | 50 | 1 | 4 |
| `...-feature-store-hash-1M-entities-300-features` | 4 | 50 | 1 | 4 |

The ingest suite writes the same entities this one reads, and is sized at `-t 10`. The serving suite — the read-heavy half of the same playbook, and the one most likely to be quoted as a serving-throughput number — runs at `-t 4` and saturates.

**Expected behavior**

Raise the serving suite's client sizing so memtier stays clear of saturation (`-t 8`–`10` with `resources.requests.cpus` matched, as every other suite in the repo does). `-c` can come down proportionally if total connection count should stay similar.

Worth reviewing `...-1M-entities-300-features` at the same time — same `-t 4`, and a 300-field row is heavier per operation, so it may be closer to the limit than the 50-field variant.

**To Reproduce**

1. Run the suite against a server with spare capacity, memtier co-located but on separate cores.
2. Sample memtier's own CPU (profiler, or `pidstat -p $(pgrep memtier_benchmark)`).
3. memtier sits at ~91% of `-t` threads while the server has idle capacity.

**Additional context**

Two sibling suites show the same pattern; filed separately for `session-caching-hash` and for the set/smembers suite. Happy to send a PR if maintainers agree with the direction.

PR #488, opened 2026-08-13
fix: raise client thread budget where memtier itself was the bottleneck

### LINKED ISSUE #485: session-caching-hash suites are load-generator bound (memtier at 91% of its threads); inconsistent with STRING/JSON siblings
**Describe the bug**

`memtier_benchmark-playbook-session-caching-hash-100k-sessions` is **load-generator bound**: memtier itself is the bottleneck, not the server under test. The suite therefore under-reports server capability and cannot detect server-side improvements.

Measured on a multi-core host with the load generator pinned to cores separate from the server:

| | |
|---|---|
| memtier CPU | **3.65 of its 4 threads — 91%** |
| server-side CPU | ~30% idle capacity remaining |

At 91% of its own thread budget the client cannot offer more load, so the reported ops/sec describes memtier's capacity rather than the server's.

**It is also inconsistent with its own family**

The session-caching playbook suites do not agree on client sizing:

| Suite | `-t` | `-c` | `resources.requests.cpus` |
|---|--:|--:|--:|
| `...-session-caching-string-100k-sessions` | **8** | 25 | 8 |
| `...-session-caching-json-100k-sessions` | **8** | 25 | 8 |
| `...-session-caching-hash-100k-sessions` | *default (4)* | *default (50)* | 4 |
| `...-session-caching-hash-100k-sessions-pipeline-16` | *default (4)* | *default (50)* | 4 |

The STRING and JSON variants model the same workload — 100k sessions, same read:write shape, same value sizes — and explicitly request `-t 8 -c 25`. The HASH variants inherit memtier's defaults. Nothing in the YAML explains the difference, and the measurement above suggests the defaults are simply too small for this workload.

**Expected behavior**

Give the HASH variants the same client sizing as their STRING/JSON siblings (`-t 8 -c 25`, `resources.requests.cpus: 8`), or whatever value keeps memtier comfortably below saturation. `resources.requests.cpus` tracks `-t` exactly across every suite in this repo, so both should move together.

The `-pipeline-16` variant is worth checking at the same time — deeper pipelining reduces client CPU per operation, so it may already have more headroom.

**To Reproduce**

1. Run the suite against any server with spare capacity, with memtier co-located but on separate cores.
2. Sample memtier's own CPU (e.g. from a profiler, or `pidstat -p $(pgrep memtier_benchmark)`).
3. memtier sits at ~91% of `-t` threads while the server has idle capacity.

**Additional context**

Happy to send a PR aligning the HASH variants with the STRING/JSON client config if maintainers agree with the direction. Related: two sibling suites show the same pattern — filed separately for the feature-store serving suite and the set/smembers suite.

fcostaoliveira:
Fixes #485, #486, #487.

## Problem

Three suites were measured genuinely load-generator bound on real hardware (server co-located, but on separate cores from the client):

| Suite | memtier CPU (of its thread budget) | server idle |
|---|---|---|
| `session-caching-hash-100k-sessions` | 91% | ~30% |
| `feature-store-hash-10M-entities-50-features` | 91% | ~33% |
| `1key-set-10-elements-smembers-pipeline-10` | 81% | ~37% |

In each case the reported ops/sec was partly a property of memtier's own capacity rather than the server under test.

## Fix

Applies the pattern already proven in #483: spread the same total connection count over more client threads (`-c 50 -t 4` → `-c 25 -t 8`, or add it explicitly where a suite relied on memtier's defaults), with `resources.requests.cpus` matched to `-t` so the extra threads aren't self-throttled by their own cpuset.

Also fixes `feature-store-hash-1M-entities-300-features` — #486's "worth reviewing at the same time" note (same `-t 4`, heavier per-op cost than the 50-features sibling).

## What was measured and deliberately left alone

Two suites in the same families were checked and are **not** changed here, because measurement didn't support it:
- `session-caching-hash-100k-sessions-pipeline-16` — 40-48% of its client thread budget. Deeper pipelining already gives it enough headroom, confirming the hypothesis in #485's own text.
- `playbook-realtime-analytics-membership` and its `-pipeline-10` variant — 29-33%. Comfortable, no fix needed.

## On #487 specifically

That issue's own text flags it as the weaker case (no in-family precedent, 81% vs. the other two suites' 91%). Local measurement on a smaller box didn't fully reproduce the reported 81% (came in around 40-51% depending on network path), but the suite was consistently the highest-loaded of the three checked in that family, its `--pipeline 10` design is directionally throughput-saturating per this repo's own memtier design guide, and the fix is the same low-risk pattern already proven elsewhere — so it's included, with medium confidence, matching the issue's own "reported for completeness, let maintainers decide" framing.

---

> [!NOTE]
> **Low Risk**
> YAML-only benchmark client tuning; no Redis or application runtime changes, though published ops/sec may shift for these suites.
> 
> **Overview**
> Updates **five memtier benchmark suites** so the client is less likely to cap reported throughput before Redis does. **Client args** move from **`-c 50 -t 4`** (or implicit defaults on the SMEMBERS pipeline case) to **`-c 25 -t 8`**, keeping the same total connection count while spreading work across more threads. **Client CPU requests** rise from **`4` to `8`** so those threads are not throttled by the pod cpuset.
> 
> Affected specs: **session-caching 100k sessions**, **feature-store** (10M/50-features main + template, and 1M/300-features), and **1-key SMEMBERS pipeline-10**. Workloads, datasets, and command mixes are unchanged—only load-generator sizing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb3cea1b6d22b4312cf1b78a516a0c41c5312478. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-set-10-elements-smembers-pipeline-10.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-set-10-elements-smembers-pipeline-10.yml
@@ -29,10 +29,10 @@ build-variants:
 clientconfig:
   run_image: redislabs/memtier_benchmark:edge
   tool: memtier_benchmark
-  arguments: --pipeline 10 --command="SMEMBERS set:10"  --hide-histogram --test-time
+  arguments: -c 25 -t 8 --pipeline 10 --command="SMEMBERS set:10"  --hide-histogram --test-time
     180
   resources:
     requests:
-      cpus: '4'
+      cpus: '8'
       memory: 2g
 priority: 23
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-playbook-feature-store-hash-10M-entities-50-features-template.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-playbook-feature-store-hash-10M-entities-50-features-template.yml
@@ -79,7 +79,7 @@ build-variants:
 clientconfig:
   run_image: redislabs/memtier_benchmark:edge
   tool: memtier_benchmark
-  arguments: '--data-size-range=8-24 --random-data --key-prefix "feature_store:user_features:entity:" --distinct-client-seed --pipeline=1 --print-percentiles=50,99 --test-time=120 -c 50 -t 4 --key-minimum=1
+  arguments: '--data-size-range=8-24 --random-data --key-prefix "feature_store:user_features:entity:" --distinct-client-seed --pipeline=1 --print-percentiles=50,99 --test-time=120 -c 25 -t 8 --key-minimum=1
     --key-maximum=10000000 --command=''HGETALL __key__'' --command-key-pattern=Z --command-ratio=19 --command=''HSET __key__ feature_1 __data__ feature_2 __data__ feature_3 __data__ feature_4 __data__ feature_5
     __data__ feature_6 __data__ feature_7 __data__ feature_8 __data__ feature_9 __data__ feature_10 __data__ feature_11 __data__ feature_12 __data__ feature_13 __data__ feature_14 __data__ feature_15 __data__
     feature_16 __data__ feature_17 __data__ feature_18 __data__ feature_19 __data__ feature_20 __data__ feature_21 __data__ feature_22 __data__ feature_23 __data__ feature_24 __data__ feature_25 __data__
@@ -90,6 +90,6 @@ clientconfig:
     '
   resources:
     requests:
-      cpus: '4'
+      cpus: '8'
       memory: 2g
 priority: 38
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-playbook-feature-store-hash-10M-entities-50-features.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-playbook-feature-store-hash-10M-entities-50-features.yml
@@ -130,8 +130,8 @@ clientconfig:
     --pipeline=1
     --print-percentiles=50,99
     --test-time=120
-    -c 50
-    -t 4
+    -c 25
+    -t 8
     --key-minimum=1
     --key-maximum=10000000
     --command='HGETALL __key__'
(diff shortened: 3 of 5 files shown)


fcostaoliveira:
Ran this through a 7-lens adversarial review before merge. Findings and what changed:

**Fixed:** also bumped `feature-store-hash-10M-entities-50-features-template.yml` — the template-encoded sibling of the suite this PR already fixes (same 10M-entity serving workload, same 19:1 HGETALL:HSET mix, differs only in field-encoding scheme). It was still at the same `-c 50 -t 4` default and would show identical saturation.

**Fresh evidence, since the original numbers were the source issues' own:** re-measured on isolated cores (server on `--io-threads 4`, client cpuset-pinned separately) against a real preloaded dataset. Old config: 362% of its 400% (4-core) budget — 90.5%, matching the ~91% #485/#486 reported. New config (`-c 25 -t 8`, cpuset expanded to 8 cores): 432% of 800% — 54%, clear headroom. Same methodology confirms the fix works, not just the diagnosis.

**Not changed, flagged instead:**
- `session-caching-hash-100k-sessions`'s `HSET` command has 7 separate `__key__` occurrences in one command — each consumes the next iterator value independently, so several fields (`organizationId`, `lastActivity`, `ipAddress`, `createdAt`) don't actually match the session's own id. Pre-existing, byte-for-byte unchanged by this PR — filing separately rather than scope-creeping this one.
- Formatting: two of the four originally-changed files insert `-c 25 -t 8` as a new first line, the other two substitute in place mid-block. Cosmetic, not fixing.

**Why not fix every suite still on `-t 4` defaults?** This PR only touches suites individually profiled and confirmed load-generator-bound. Applying `-t 8` blanket-wide risks adding client overhead to suites that were never client-constrained. The systemic fix — the runner warning when memtier's own CPU crosses a saturation threshold during a run, proposed in #487 — is the right way to surface the rest; that belongs in a follow-up issue, not a manual audit here.

PR #514, opened 2026-08-18
fix(test-suites): correct zset cardinality claims and two independent hit-rate bugs (#508)

### LINKED ISSUE #508: Shared zset preload (ZADD lb __key__  __key__) undercounts cardinality due to per-occurrence key draws
## Summary

`ZADD lb __key__  __key__` is the shared preload command reused verbatim by several
`1key-zset-*-elements-integer` dataset specs. It uses **two** `__key__` placeholders
in one command (intending `score == member`), but memtier evaluates every `__key__`
occurrence in a command as its own **independent** random/sequential draw, not one
shared value. Combined with the `P` (sequential/partitioned) key-pattern, each preload
op consumes **two** slots from the pattern's counter instead of one, so the counter
wraps and overwrites partway through the run -- the resulting zset ends up with
**fewer distinct elements than the nominal dataset size** (and score != member for
most entries).

## Evidence

Verified empirically (local `redis-server` + `memtier_benchmark:edge` docker,
preload command copied verbatim from `memtier_benchmark-1key-zrank-1M-elements-pipeline-1.yml`
scaled down 1:4000 for a fast repro):

```
--pipeline 50 --command="ZADD lb __key__  __key__" --command-key-pattern=P \
  --key-maximum 250 --key-prefix "" -n 25 --hide-histogram -t 10 -c 1
```
(nominal 25*10*1 = 250 total ops, `--key-maximum 250`)

Result: `ZCARD lb` = **134**, not 250 (~54% of the nominal count). The same relative
oversubscription ratio (2 placeholders * total-ops / (key-maximum+1)) applies
identically to the 1M-element preload (`-n 100000 -t 10 -c 1`, `--key-maximum
1000000`), so the "1 million elements" claim in these specs' `dataset_description`
is almost certainly similarly overstated.

## Affected specs (share the `1key-zset-1M-elements-integer` dataset_name / preload)

- `memtier_benchmark-1key-zrank-1M-elements-pipeline-1.yml`
- `memtier_benchmark-1key-zrevrank-1M-elements-pipeline-1.yml`
- `memtier_benchmark-1key-zincrby-1M-elements-pipeline-1.yml`
- `memtier_benchmark-1key-zset-1M-elements-zcard-pipeline-10.yml`
- `memtier_benchmark-1key-zset-1M-elements-zscore-pipeline-10.yml`
- `memtier_benchmark-1key-zset-1M-elements-zremrangebyscore-pipeline-10.yml`
- `memtier_benchmark-1key-zset-1M-elements-zrevrange-5-elements.yml`
- `memtier_benchmark-1key-zset-1M-elements-zrevrange-withscores-5-elements-pipeline-10.yml`

(Found while investigating #505, which reuses this same preload pattern for its two
new 1M-element specs -- those are unaffected in *purpose* since their new-member
namespace (`newmember__key__`) can't collide with the bare `__key__` preload members,
but their `dataset_description` inherits the same "1M elements" overstatement.)

## Suggested fix

Change the preload command to use a single `__key__` placeholder reused via a
literal, e.g. `ZADD lb __key__ member__key__` won't fix the *score* side (still an
independent draw) -- the cleanest fix is likely switching preload to a plain
single-field command per op (e.g. two passes, or `--command-key-pattern=P` with a
command that only needs one placeholder) so score and member are guaranteed to
derive from the same counter value. Needs design thought; filing to track rather
than proposing a specific patch here.

`check: keyspacelen: 1` only validates key *count*, not zset cardinality, so this
has never failed CI.

fcostaoliveira:
## Summary

Fixes #508 by correcting the `1 million elements` claim across 8 specs sharing the `1key-zset-1M-elements-integer` dataset (preload: `ZADD lb __key__  __key__`, P key-pattern) to the true, empirically-verified cardinality.

The preload's two `__key__` occurrences are each independently drawn by memtier's P pattern rather than sharing one value, so real cardinality falls short of the nominal 1,000,000 by an amount that depends on each spec's concurrency shape (verified live per shape):

| Shape | Files | Real cardinality |
|---|---|---|
| `-n 100000 -t 10 -c 1` | zrank, zadd-new-random-members, zrangebyscore-random-position | 949,996 |
| `-t 4 -c 100` (default `-n`=10000) | zscore, zremrangebyscore, zcard, zrevrange x2 | 998,950 |
| `-n 1000000 -t 1 -c 1` | zrevrank, zincrby | 1,000,000 (accurate as-is, unchanged) |

**Why description-only, not a preload fix:** a domain-widening fix (the approach used for the still-unmerged #512/#513 specs, which had no shipped history) would regress these specs' hit rate from ~95%+ down to ~50%, since real members would become a sparse even-only sub-range that memtier's client-side `R` pattern can't selectively target. Correcting the documentation to match reality is the safe fix here; #508's own text flagged this needs design thought rather than a copy-paste.

## Two more severe bugs found on adjacent files

While verifying #508's claim empirically, found two independent, more severe defects on files in the same dataset family:

- **`zscore-pipeline-10`**: client omitted `--key-maximum`, defaulting to memtier's own `10,000,000` range against a ~1M-element dataset. Confirmed live: **89.72% miss rate**. Fixed by adding `--key-maximum 1000000`; re-verified at 99.87% hit rate.
- **`zremrangebyscore-pipeline-10`**: client omitted **both** `--key-maximum` and `--key-prefix`. Without `--key-prefix`, memtier defaults substituted values to a `"memtier-"` prefix (e.g. `"memtier-138344"`), so every call sent non-numeric score bounds. Confirmed live via `MONITOR` and zero net `ZCARD` change across hundreds of thousands of calls: **this spec has never removed a single real element** -- it only measured Redis's invalid-float error-reply latency. Fixed both flags. Documented in the description that, once functional, the wide-range random-draw design still drains the set to <1% within ~100 calls (verified live), so most of the run measures the empty-range fast path -- a proper steady-state fix needs a producer/refill design (tracked as a follow-up, out of scope here).

## Verification

- `redis-benchmarks-spec-cli --tool stats --fail-on-required-diff`: exit 0
- `pytest utils/tests/test_spec.py utils/tests/test_scope.py`: 27/27 passed
- `black --check`: clean
- All cardinality/hit-rate numbers above verified live against a real `redis-server` + `redislabs/memtier_benchmark:edge`, not derived from reasoning alone (multiple earlier assumptions in this investigation turned out wrong empirically).

Fixes #508

---

> [!NOTE]
> **Medium Risk**
> YAML-only, but fixing ZSCORE and ZREMRANGEBYSCORE client args changes real benchmark workloads and will shift reported performance versus prior broken runs.
> 
> **Overview**
> Updates **eight** memtier specs on the `1key-zset-1M-elements-integer` dataset so descriptions and `dataset_description` state **ZCARD-verified** sizes (~949,996 or ~998,950) instead of a nominal 1M, and document the shared preload quirk where two `__key__` tokens under pattern `P` get independent draws ([#508](https://github.com/redis/redis/issues/508)).
> 
> **`zscore-pipeline-10`** now passes `--key-maximum 1000000` on the client so lookups target the loaded score range (~90% misses before).
> 
> **`zremrangebyscore-pipeline-10`** adds `--key-maximum 1000000` and `--key-prefix ""` so score bounds are numeric (previously default `memtier-` prefixes made every call an invalid-score error with no removals); the description also notes the spec still drains the set quickly once removals work, with steady-state refill left as follow-up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9d8894a335f3c260dc12caba49d529b833048f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zrank-1M-elements-pipeline-1.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zrank-1M-elements-pipeline-1.yml
@@ -1,8 +1,13 @@
 version: 0.4
 name: memtier_benchmark-1key-zrank-1M-elements-pipeline-1
 description: 'Runs memtier_benchmark, for a keyspace length of 1 SORTED SET key. The SORTED SET
-  contains 1M elements and we query it using ZRANK. Encoding: skiplist+hashtable (1,000,000
-  elements, exceeds zset-max-listpack-entries 128).'
+  contains 949,996 elements (verified via ZCARD), not the nominal 1,000,000 -- the shared
+  preload command (ZADD lb __key__ __key__) has two __key__ occurrences, and memtier''s P
+  key-pattern draws a fresh sequential value per occurrence rather than one shared value, so
+  cross-connection wraparound leaves the set ~5% short of the nominal count (see #508). Query
+  member lookups (ZRANK) still hit real, densely-packed members at a ~95%+ rate since the
+  populated range still spans close to the full nominal domain. Encoding: skiplist+hashtable
+  (949,996 elements, exceeds zset-max-listpack-entries 128).'
 dbconfig:
   configuration-parameters:
     save: '""'
@@ -17,8 +22,8 @@ dbconfig:
     requests:
       memory: 1g
   dataset_name: 1key-zset-1M-elements-integer
-  dataset_description: This dataset contains 1 sorted set key with 1 million elements,
-    each with an integer score.
+  dataset_description: This dataset contains 1 sorted set key with 949,996 elements (verified
+    via ZCARD, not the nominal 1,000,000 -- see #508), each with an integer score.
 tested-groups:
 - sorted-set
 tested-commands:
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zset-1M-elements-zadd-new-random-members.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zset-1M-elements-zadd-new-random-members.yml
@@ -1,9 +1,9 @@
 version: 0.4
 name: memtier_benchmark-1key-zset-1M-elements-zadd-new-random-members
 description: 'Runs memtier_benchmark, for a keyspace length of 1 SORTED SET key. The SORTED SET
-  is pre-loaded with a nominal 1M elements (integer scores drawn from the 1..1,000,000 range --
-  the actual populated count is somewhat lower than 1M due to a separate, pre-existing preload
-  quirk tracked in #508) then we ZADD members named newmember__key__ with a score, where the
+  is pre-loaded with 949,996 elements (verified via ZCARD, not the nominal 1,000,000 -- the
+  shared preload has a dual-__key__-draw mechanic documented in #508 and its fix) then we ZADD
+  members named newmember__key__ with a score, where the
   member-name suffix and the score are each drawn independently and uniformly at random from that
   SAME 1..1,000,000 range -- memtier evaluates every __key__ occurrence in a command as its own
   independent draw, not one shared value, so the score is not simply "the member''s number" (the
@@ -34,8 +34,8 @@ dbconfig:
     requests:
       memory: 1g
   dataset_name: 1key-zset-1M-elements-integer
-  dataset_description: This dataset contains 1 sorted set key with 1 million elements,
-    each with an integer score.
+  dataset_description: This dataset contains 1 sorted set key with 949,996 elements (verified
+    via ZCARD, not the nominal 1,000,000 -- see #508), each with an integer score.
 tested-groups:
 - sorted-set
 tested-commands:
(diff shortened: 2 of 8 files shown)


fcostaoliveira:
Filed #515 to track the deeper zremrangebyscore-pipeline-10 design limitation (drains to a low single-digit percent of the set within ~100 calls, then measures the empty-range fast path for the rest of the run) -- confirmed a producer/refill multi-clientconfig redesign (matching #442's pattern) is needed, which is out of scope for this PR's bug fixes.

fcostaoliveira:
9-lens adversarial review, on head d9d8894a335f3c260dc12caba49d529b833048f1:

All 9 lenses independently re-verified live (fresh redis-server + memtier_benchmark:edge), no unresolved BLOCKING findings:
- correctness: independently reproduced the zscore 89.7%->99.87% hit-rate fix and the zremrangebyscore non-numeric-bounds bug (confirmed via MONITOR), plus spot-checked cardinality numbers.
- description-accuracy: found and fixed one leftover stale '1,000,000' mention in zadd-new-random-members.yml (d9d8894); re-scanned all 8 files, clean now.
- scope-and-design-judgment: the docs-only fix for the 6 pure-cardinality files is justified (widening the domain would regress hit rate to ~50%); filed #515 to track zremrangebyscore's remaining drain-to-empty limitation rather than leaving it undocumented.
- regression-check: zrevrank/zincrby confirmed untouched (zero diff vs main), consistent with their shape already being accurate.
- test-coverage: stats --fail-on-required-diff exit 0, pytest utils/tests/test_spec.py utils/tests/test_scope.py 27/27 passed, black clean.
- style/security: clean, consistent with sibling specs.
- CI-status: 9/9 green (Stream field contract, Validate SPEC fields, Cursor Bugbot, all 6 lint/test matrix jobs).
- maintainer-appetite: mergeable, blocked only by GitHub's self-approval restriction on my own PR.

PR #522, opened 2026-08-22
fix: raise docker-py client default timeout to avoid spurious ReadTimeout under load

fcostaoliveira:
## Summary
- `docker.from_env()` with no explicit `timeout` falls back to docker-py's own `DEFAULT_TIMEOUT_SECONDS` (60s), which applies to every API call that doesn't set its own override -- container logs/inspect, not just `container.wait()` (which already computes a generous per-call timeout in `run_client_configs()`).
- On a busy shared Docker host, an ordinary fast call can still exceed 60s while the daemon is occupied handling other concurrent containers, producing a `ReadTimeout` even though the container itself is healthy.
- Raises the client-wide default from 60s to 300s at every `docker.from_env()` call site (builder, runner, self-contained coordinator, cli).

## Motivation
Observed intermittent `UnixHTTPConnectionPool(host='localhost', port=None): Read timed out` failures on otherwise-healthy benchmark runs, on two distinct call paths:
- A long-running blocking-client spec, via `container.wait()` -- its own per-call timeout is `test-time + 60s` buffer, comfortably long enough in isolation, but the failure still occurred, consistent with host-level Docker daemon contention rather than the container's own runtime.
- A short, ordinary (non-blocking) spec, via a plain container inspect call that has no per-call timeout override at all and was therefore governed purely by the 60s client default.

Both are consistent with the same root cause: the client-wide default timeout is too short for a Docker host under concurrent load, for API calls that don't set their own override. This is a client-side connection-timeout floor, not a change to any per-call timeout logic (`container.wait()`'s own computed timeout, `container_timeout = test_time + timeout_buffer`, is untouched and still wins for that call).

## Test plan
- [x] All 4 changed files parse cleanly (`ast.parse`)
- [x] `docker.from_env(timeout=300)` constructs successfully against a local Docker daemon
- [x] Confirmed via `docker-py` source (`docker/api/container.py`, `container.wait()`) that a per-call `timeout` argument always takes precedence over the client default, so specs with `--test-time` up to 600s (this repo's longest) are unaffected by this change
- [x] Checked existing tests (`utils/tests/test_*.py`) construct their own `docker.from_env()` fixtures rather than asserting on this repo's call arguments, so the added kwarg doesn't affect their expectations

---

> [!NOTE]
> **Low Risk**
> Timeout-only change to Docker client construction; per-call timeouts (e.g. `container.wait()`) still override the default. No auth, data, or benchmark-duration logic is modified.
> 
> **Overview**
> Raises the docker-py client-wide default timeout from **60s to 300s** at every `docker.from_env()` call (builder, runner, self-contained coordinator, CLI).
> 
> The 60s default applies to any API call without its own override (inspect/logs/list). On a contended shared Docker daemon those calls can exceed 60s even when the container is healthy, causing `UnixHTTPConnectionPool` `ReadTimeout` failures (seen on BZPOPMIN and ARM ZREMRANGEBYSCORE runs). Per-call timeouts such as `container.wait()` still take precedence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d71346b8acb4e64bb90cead54c57361abd757b65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/__builder__/builder.py
+++ b/redis_benchmarks_specification/__builder__/builder.py
@@ -360,7 +360,11 @@ def builder_process_stream(
         # commit = None
         # commited_date = ""
         # tag = ""
-        docker_client = docker.from_env()
+        # Explicit timeout: docker-py's client-wide default is 60s, applied to
+        # every API call that doesn't pass its own override. A busy shared
+        # Docker host can exceed 60s for an ordinary, fast call, so a longer
+        # client-wide default avoids spurious ReadTimeout failures under load.
+        docker_client = docker.from_env(timeout=300)
         from pathlib import Path
 
         build_request_arch = None
@@ -1094,7 +1098,11 @@ def build_spec_image_prefetch(builders_folder, different_build_specs):
     logging.info("checking build spec requirements")
     already_checked_images = []
     hub_pulled_images = 0
-    client = docker.from_env()
+    # Explicit timeout: docker-py's client-wide default is 60s, applied to
+    # every API call that doesn't pass its own override. A busy shared Docker
+    # host can exceed 60s for an ordinary, fast call, so a longer client-wide
+    # default avoids spurious ReadTimeout failures under load.
+    client = docker.from_env(timeout=300)
     for build_spec in different_build_specs:
         build_config, id = get_build_config(builders_folder + "/" + build_spec)
         if build_config["kind"] == "docker":
--- a/redis_benchmarks_specification/__cli__/cli.py
+++ b/redis_benchmarks_specification/__cli__/cli.py
@@ -283,7 +283,13 @@ def trigger_tests_dockerhub_cli_command_logic(args, project_name, project_versio
         build_stream_fields["target_platform"] = args.target_platform
         logging.info(f"Targeting specific runner: {args.target_platform}")
     if args.docker_dont_air_gap is False:
-        docker_client = docker.from_env()
+        # Explicit timeout: docker-py's client-wide default is 60s, applied to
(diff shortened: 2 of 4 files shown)


fcostaoliveira:
9-lens adversarial review, on head d71346b8acb4e64bb90cead54c57361abd757b65:

All 9 lenses approve, no BLOCKING findings:
- correctness: all 8 production docker.from_env() call sites correctly pass timeout=300; grepped the whole repo, confirmed no site missed (only out-of-scope test fixtures remain untouched). Verified live: docker.from_env(timeout=300).api.timeout == 300.
- precedence-claim-verification: read docker-py source (APIClient._set_request_timeout uses kwargs.setdefault('timeout', self.timeout)) -- an explicit container.wait(timeout=N) always wins over the client default. Confirmed the caller's container_timeout is always a real number, never None, so this holds in actual usage.
- side-effects: no retry/error-handling code in this codebase assumes a ~60s failure window; the separate stuck-container staleness check is independent of the HTTP client timeout.
- test-coverage: all relevant test files pass (78/78 across builder/cli/runner/coordinator tests, excluding 2 pre-existing hangs confirmed present on main without this change too -- not a regression).
- scope-completeness: 300s is fine since .wait()'s own per-call timeout governs the longest specs independently of the client-wide default.
- style/security: clean, consistent formatting across all 8 sites.
- CI-status: 9/9 green, zero Bugbot findings.
- maintainer-appetite: motivation is concrete and independently verifiable (names two specific prior benchmark runs from this session).

Issue #526, opened 2026-08-22
[SPEC]: variadic generic commands (DEL/UNLINK/EXISTS/TOUCH) benchmarked only at width 1; MSETNX has no coverage


fcostaoliveira:
The variadic generic commands are benchmarked only at **one key**, which is the width at which they stop being variadic.

| Command | Suites | Widths sampled |
|---|---|---|
| `MGET` | 18 | 1, 2, 5, 10, 20, 30, 100 |
| `MSET` | 7 | 3, 10, 50 |
| `DEL` | 6 | **1** |
| `UNLINK` | 1 | **1** |
| `EXISTS` | 1 | **1** |
| `TOUCH` | 1 | **1** |
| `MSETNX` | **0** | — |

`MGET` and `MSET` are in good shape. The generic four are not: every suite issues them as `--command "DEL __key__"`, `--command "EXISTS __key__"`, `--command "TOUCH __key__"`. Per-command cost is roughly `C_fixed + N * C_key`, and at N=1 only `C_fixed` is observable — so argument-vector growth, the multi-key lookup loop, per-key free versus lazy free, and the tail-latency cost of one command occupying the event loop are all unmeasured.

### Proposal — 7 suites on a 1 / 10 / 100 decade

Widths 10 and 100 are already the established `MGET`/`MSET` points, so these become controlled comparisons rather than a new axis.

- `DEL` at 10 and 100 — per-key free dominates at 100
- `UNLINK` at 100 — **paired with `DEL` at identical width**, which is the only way to price the async-free path against synchronous free; at width 1 the two are indistinguishable
- `EXISTS` at 100 — read-only, no free: isolates pure N-lookup from `DEL`'s free cost
- `TOUCH` at 100 — same shape as `EXISTS` but performs LRU/LFU and expire maintenance per key, so `TOUCH - EXISTS` prices that maintenance at a width where it resolves
- `MSETNX` at 10 — currently zero coverage; pairs with the existing `MSET` at 10, since `MSETNX` is `MSET` plus N existence pre-checks
- `MSET` at 100 — completes the write side at `MGET`'s widest point

Two notes on correctness, both verified against the existing corpus:

[...]

PR #529, opened 2026-08-22
Add 100-key variadic coverage for EXISTS, TOUCH and MSET

fcostaoliveira:
## Summary

`EXISTS`, `TOUCH`, `DEL` and `UNLINK` are variadic, but every suite that exercises them passes exactly **one** key:

```
--command "EXISTS __key__"
--command "TOUCH __key__"
```

At width 1 only the fixed per-command cost is observable. The per-key lookup loop, argument-vector growth, and the tail-latency cost of a single command occupying the event loop are all invisible. `MGET` already spans widths 1-100 and is in good shape; the generic commands are not.

This PR adds three suites at 100 keys per call:

| Suite | Width |
|---|---|
| `memtier_benchmark-1Mkeys-generic-exists-100keys` | 100 |
| `memtier_benchmark-1Mkeys-generic-touch-100keys` | 100 |
| `memtier_benchmark-5Mkeys-string-mset-100B-100keys` | 100 pairs |

### EXISTS and TOUCH are a matched pair

Both perform N lookups; `TOUCH` additionally updates LRU/LFU metadata per key. The difference between the two isolates that maintenance cost at a width where it resolves — so they are deliberately identical in every respect except the verb: same dataset (`1Mkeys-string-100B-size`), same `--key-maximum`, same connection counts, same pipeline depth, byte-identical argument strings otherwise.

Worth noting they would **not** have been comparable if derived naively from the two existing single-key suites — those use different datasets (hash vs string-with-expiration) and different key ranges, so the difference would have conflated data type, TTL presence, and hit ratio with the LRU cost.

`--key-maximum` is twice the preloaded keyspace, so each call mixes hits and misses within a single command rather than measuring an all-hit best case.

### Not included

`DEL` and `UNLINK` at width belong in a separate PR: a width-100 delete consumes keys ~100x faster than a single-key suite, so those need the replenishing-writer pattern from `memtier_benchmark-1Mkeys-string-key-churn-rehash-32B` and a validity criterion on the miss ratio. Worth reviewing on its own.

`MSETNX` is also held back. Against a populated keyspace it always returns 0 and writes nothing — it still performs all N existence checks, but a suite whose command never succeeds needs a deliberate design rather than being folded in here.

Refs #526.

## Verification

```
  EXIT=0
  12 warnings — unchanged from the pre-existing baseline, none referencing the new suites

  40 passed in 23.15s

placeholder counts asserted programmatically: exists 100 __key__, touch 100 __key__,
mset 100 __key__ + 100 __data__
```

`tested-groups` taken from the validator's derived list, not hand-written.

---

> [!NOTE]
> **Low Risk**
> Adds YAML benchmark specs only; no runtime, auth, or data-handling code. New suites will increase CI/benchmark load.
> 
> **Overview**
> Adds three memtier test suites that run **100 keys per command**, filling a gap where existing `EXISTS`/`TOUCH` suites used a single key and `MSET` topped out at 50 pairs.
> 
> `EXISTS` and `TOUCH` share the same 1M-key RAW string dataset, key range (hits mixed with misses via `--key-maximum` 2M), pipeline, and client shape so the delta isolates LRU/LFU maintenance. `MSET` uses a 5M-key 100B dataset to match existing 100-key `MGET` width; measured values use constant filler to focus on dispatch/arg-vector overhead.
> 
> Standalone-only, gcc amd64/arm64 plus dockerhub. `DEL`/`UNLINK`/`MSETNX` at this width are intentionally omitted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e7ee64c29700c8a7a6bdc5d0a338226c48769ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-generic-exists-100keys.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-generic-exists-100keys.yml
@@ -1,0 +1,48 @@
+version: 0.4
+name: memtier_benchmark-1Mkeys-generic-exists-100keys
+description: Runs memtier_benchmark issuing EXISTS with 100 keys per call against a 1M-key dataset. EXISTS is variadic
+  but every existing suite calls it with a single key, where only the fixed per-command cost is visible; at 100
+  keys the per-key lookup loop dominates. --key-maximum is twice the preloaded keyspace, so each call mixes hits
+  and misses rather than measuring an all-hit best case.
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  check:
+    keyspacelen: 1000000
+  preload_tool:
+    run_image: redislabs/memtier_benchmark:edge
+    tool: memtier_benchmark
+    arguments: '"--data-size" "100" --command "SET __key__ __data__" --command-key-pattern="P" -n 5000 --key-minimum=1
+      --key-maximum 1000000 -c 50 -t 4 --hide-histogram --pipeline 50'
+  resources:
+    requests:
+      memory: 1g
+  dataset_name: 1Mkeys-string-100B-size
+  dataset_description: This dataset contains 1 million string keys, each with a data size of 100 bytes.
+tested-commands:
+- exists
+tested-groups:
+- generic
+redis-topologies:
+- oss-standalone
+build-variants:
+- gcc:15.2.0-amd64-debian-bookworm-default
+- gcc:15.2.0-arm64-debian-bookworm-default
+- dockerhub
+clientconfig:
+  run_image: redislabs/memtier_benchmark:edge
+  tool: memtier_benchmark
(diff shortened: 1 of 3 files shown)


fcostaoliveira:
Good catch, and it was a real one — fixed in 
`c66c786`.

The suite was derived from the MGET 100-key suite, which is read-only and so carries `--data-size` only in its preload. MSET writes, so memtier would have fallen back to its 32-byte default: ~32B values under a suite named and described as 100B, on the wrong side of the EMBSTR/RAW boundary, and not comparable to the MGET 100B suite this exists to pair with.

Both existing MSET suites set `--data-size 100` on the client; this now matches them. Validator still exits 0 with the warning count unchanged.

Issue #525, opened 2026-08-22
[SPEC]: numkeys family — 8 commands with zero coverage; the 2 existing suites are pinned at the minimum width


fcostaoliveira:
The `numkeys` command family — where the key count is a runtime integer argument rather than a fixed arity — is effectively unbenchmarked. Coverage today:

| Command | Suites |
|---|---|
| `LMPOP`, `ZMPOP`, `SINTERCARD`, `ZINTERCARD`, `ZDIFF`, `ZINTER`, `ZDIFFSTORE`, `ZINTERSTORE` | **0** |
| `BLMPOP`, `BZMPOP` | **0** |
| `ZUNION`, `ZUNIONSTORE` | 1 each |

### The two existing suites don't exercise the family's defining property

Both pin `numkeys` to **2** — the minimum legal value, and the one width at which the numkeys loop is indistinguishable from a fixed 2-arity command:

```
ZUNION 2 zset:skiplist:1:300 zset:skiplist:2:300
ZUNIONSTORE zset:skiplist:3:600 2 zset:skiplist:1:300 zset:skiplist:2:300
```

There is no sibling suite at any other width, so the variadic axis is unmeasured across all twelve commands.

### How much is being missed

Measured on a local Redis, identical 500-element sorted sets, `-c 1 -t 1 --pipeline 1 --test-time 3`:

| `ZINTERCARD` numkeys | ops/sec |
|---:|---:|
| 2 | 9,892 |
| 8 | 2,208 |
| 64 | 356 |

**~28x** across the one dimension no suite varies. A regression anywhere in the key-collection loop, the argument-vector allocation, or the per-key lookup path would be invisible at width 2 and obvious at width 64.

### Proposal

A small, dataset-shared set rather than one suite per command per width:

**Priority 1 — the width axis plus the zero-coverage commands (8 suites)**

- `ZINTERCARD` at widths **2 / 8 / 64** — establishes the curve; the other commands then need only one point each
- `SINTERCARD` at width 8 — the set-side parser, separate implementation
- `ZDIFF` at width 8 over disjoint sets, so the reply is non-empty
- `ZINTER` at width 8 — reply-producing counterpart to `ZINTERCARD`
- `LMPOP` / `ZMPOP` over 8 keys where only the last is populated — every call performs all 8 lookups

[...]

PR #531, opened 2026-08-22
Add ZINTER, ZDIFF and SINTERCARD coverage at numkeys=8

fcostaoliveira:
## Summary

`ZINTER`, `ZDIFF` and `SINTERCARD` have **no suite at all** today. This adds one each at `numkeys=8`.

| Suite | Command | Reply |
|---|---|---|
| `memtier_benchmark-8keys-zset-500-elements-zinter` | `ZINTER 8 …` | 500 elements |
| `memtier_benchmark-8keys-zset-500-elements-zdiff` | `ZDIFF 8 …` | 500 elements |
| `memtier_benchmark-8keys-set-500-elements-sintercard` | `SINTERCARD 8 … LIMIT 0` | 500 |

### Seeding is the part that matters

Each dataset is chosen so the reply is **non-empty and constant**, which is less obvious than it sounds:

- **ZDIFF over an overlapping dataset returns nothing.** A suite seeded the same way as the ZINTER one would benchmark the empty-reply path rather than the diff. Its 8 member sets are therefore pairwise disjoint, so the reply is the full first set.
- **ZINTER and SINTERCARD** use identical member sets across all 8 keys, so the reply is a constant 500 regardless of width — leaving `numkeys` as the only variable if width sweeps are added later.

`SINTERCARD` is a separate implementation from the sorted-set numkeys commands, so it does not inherit coverage from them.

Seeding uses `init_lua` rather than inline `ZADD`/`SADD` literals — the two pre-existing suites in this family carry ~80KB of literals each.

Refs #525.

## Verification

```
  EXIT=0
  12 warnings — unchanged baseline, none referencing the new suites

  40 passed in 22.84s
```

`tested-groups` cross-checked programmatically against each command's `group` in `commands.json`: ZINTER/ZDIFF → `sorted-set`, SINTERCARD → `set`.

---

> [!NOTE]
> **Low Risk**
> Adds standalone benchmark YAML only; no runtime, auth, or data-path code changes.
> 
> **Overview**
> Adds first-time memtier coverage for `ZINTER`, `ZDIFF`, and `SINTERCARD` at `numkeys=8` (500 elements per key, skiplist/hashtable encodings).
> 
> Datasets are seeded via `init_lua` so replies stay non-empty and constant: identical members for `ZINTER`/`SINTERCARD` (cardinality 500), pairwise-disjoint members for `ZDIFF` so the reply is the full first set rather than empty. Suites run 120s on oss-standalone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc0ca2e15000e41974b4cc5e465c09ac8deef7db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-8keys-set-500-elements-sintercard.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-8keys-set-500-elements-sintercard.yml
@@ -1,0 +1,41 @@
+version: 0.4
+name: memtier_benchmark-8keys-set-500-elements-sintercard
+description: 'Runs memtier_benchmark issuing SINTERCARD over 8 sets of 500 elements each. SINTERCARD has no coverage today and is a separate implementation from the sorted-set numkeys commands. All 8 keys hold the identical member set so the cardinality is a constant 500.'
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  check:
+    keyspacelen: 8
+  resources:
+    requests:
+      memory: 1g
+  init_lua: |
+    -- 8 sets all holding the identical 500-member set, so the intersection cardinality
+    -- is a constant 500.
+    for k = 1, 8 do
+      for m = 1, 500 do
+        redis.call('SADD', 'st:' .. k, 'm:' .. m)
+      end
+    end
+    return 1
+  dataset_name: 8keys-set-500-elements-identical-members
+  dataset_description: 8 sets, each containing the identical 500 members m:1..m:500.
+tested-commands:
+- sintercard
+redis-topologies:
+- oss-standalone
+build-variants:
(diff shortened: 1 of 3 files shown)


fcostaoliveira:
Correct, and fixed in `54c8e3b`.

The 64-key dataset was deliberate but wrong: I wanted the preload shared with the ZINTERCARD width sweep in #528. That only pays off if both land, and it is not worth a suite whose `keyspacelen` and preload are 8x what it exercises — especially when the ZDIFF and SINTERCARD suites in this same PR both seed 8.

Now seeds 8, with `keyspacelen: 8` and its own `dataset_name`. All three suites in the PR are consistent: seeded == numkeys == keyspacelen, asserted programmatically.

PR #534, opened 2026-08-22
Add LMPOP and ZMPOP coverage at numkeys=8, bounded

fcostaoliveira:
## Summary

`LMPOP` and `ZMPOP` have **no coverage today**. This adds one bounded suite each at `numkeys=8`.

### Why only the last key is populated

`LMPOP` and `ZMPOP` scan the provided key names **in order and stop at the first non-empty one**. Populating the first key would therefore measure a single lookup and tell you nothing about `numkeys` — the suite would be named an 8-key workload while behaving like a 1-key one.

Populating only the **last** of the 8 forces all 8 lookups on every call, which is the fan-out these commands exist to do.

### Why bounded rather than time-based

Both are destructive. Against `--test-time` the collection drains and the suite silently switches to measuring the empty reply — a throughput number that describes an empty keyspace.

Each client is bounded to `-n 100000` at `-c 1 -t 1`, i.e. exactly 100,000 pops against a 500,000-element preload (**20%**), so draining is impossible by construction. This follows `memtier_benchmark-1key-zset-1M-elements-zpopmin-bounded`, including documenting the preload-domain arithmetic in the description.

Refs #525.

## Verification

```
pop budget asserted programmatically:
  lmpop  seeded=500000  pops=100000  (20% of preload)
  zmpop  seeded=500000  pops=100000  (20% of preload)

  EXIT=0 — 12 warnings, unchanged baseline, none referencing the new suites

  40 passed in 22.36s
```

`tested-groups` cross-checked against `commands.json`: LMPOP → `list`, ZMPOP → `sorted-set`.

---

> [!NOTE]
> **Low Risk**
> Adds YAML benchmark specs only; no runtime, auth, or data-handling code changes.
> 
> **Overview**
> Adds first-time coverage for `LMPOP` and `ZMPOP` via two bounded memtier suites on standalone.
> 
> Each suite preloads a single 500K-element collection on the **last** of eight key names so every call walks all `numkeys` lookups instead of stopping on the first hit. Clients run a fixed `-n 100000` (20% of preload) so destructive pops cannot drain the key and start measuring empty replies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df7e5b2d6a779a2cd12e9c0182f0e37c9b31c220. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Correct — fixed in `df7e5b2`.

I copied the shape of the `zpopmin-bounded` preload without applying its arithmetic. With two `__key__` placeholders each command draws two domain positions, so 500,000 commands need a 1,000,000 domain; at 500,000 the draws wrap and ZCARD lands below the claimed 500,000, which invalidates the 20% pop budget the whole bounded design rests on.

`--key-maximum` is now 1000000, matching the precedent's 2x ratio for the dual-placeholder case, and the description states the reasoning rather than just the number.

LMPOP is unaffected — its `LPUSH` preload has a single placeholder, so its domain was already correctly sized.

PR #535, opened 2026-08-24
Add ZMSCORE member-count coverage at 1/16/32/33 (prefetch batch boundaries)

fcostaoliveira:
The existing `memtier_benchmark-1key-zset-1M-elements-zmscore-100-members` suite only
exercises a wide member count (100), deep into multi-batch prefetch territory. It never
measures:
- **1 member** — below the batching threshold, the plain unbatched lookup loop.
- **16 members** — exactly the server-side batch-size constant (`ZMSCORE_PREFETCH_BATCH`),
  handled as a single undivided `dictPrefetchKeys()` call.
- **32 members** — the first width where batching actually splits into multiple calls: a
  clean, even 16+16.
- **33 members** — one past that split: 16+17, not 16+16+1. This is the direct proof that
  the adaptive-shrink rule folds a would-be trailing remainder into the previous batch
  rather than stranding it as a batch of 1 (which the underlying prefetch primitive
  silently gives no benefit to) — the exact shape a naive fixed-size chunker would get
  wrong.

All 4 new suites are siblings of the existing 100-member one: same 1M-element
constant-score dataset, same preload, same query shape — only the ZMSCORE member count
differs, isolating the member-count axis cleanly.

Reviewed against the memtier design/review checklist: encoding + metric stated in each
description (skiplist, Ops/sec), `keyspacelen`/preload identical to the already-verified
sibling, single-letter `--command-key-pattern`, no drain risk (read-only), no newer-build-only
flags on the `:edge` image.

(Original PR used 1/16/17 — Cursor Bugbot correctly caught that 16/17 don't actually exercise
any batch-splitting at all with the current `ZMSCORE_PREFETCH_BATCH=16` constant; replaced
with the real 32/33 boundary. See PR comments.)

---

> [!NOTE]
> **Low Risk**
> Adds YAML benchmark specs only; no runtime, auth, or data-handling code. Extra CI load from four new standalone suites.
> 
> **Overview**
> Adds four sibling `ZMSCORE` memtier suites on the same 1M-element constant-score skiplist dataset as the existing 100-member suite, isolating member count at **1 / 16 / 32 / 33**.
> 
> Those widths map to the intra-command dict-prefetch path: unbatched lookup (1), a single `ZMSCORE_PREFETCH_BATCH` of 16, the first multi-batch split (32 = 16+16), and the adaptive-shrink remainder (33 = 16+17 instead of a stranded 1). Metric remains Ops/sec; preload, topology, and build variants match the 100-member sibling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9822ba7a8d3f1a3e1d1216c6f716c031742d7601. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zset-1M-elements-zmscore-1-members.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-zset-1M-elements-zmscore-1-members.yml
@@ -1,0 +1,47 @@
+version: 0.4
+name: memtier_benchmark-1key-zset-1M-elements-zmscore-1-members
+description: 'Runs memtier_benchmark, for a keyspace length of 1 SORTED SET key. The SORTED SET
+  contains 1,000,000 elements (integer members 1..1,000,000, all with a constant score of 1) and
+  we query it using ZMSCORE with 1 independent random members per call. Sibling of the existing
+  memtier_benchmark-1key-zset-1M-elements-zmscore-100-members suite (same dataset, same preload,
+  same query shape) -- this family sweeps the member-count axis at 1/16/17/100 to cover the
+  intra-command dict-prefetch batching thresholds that the 100-member-only suite left unmeasured.
+  This width sits below the intra-command dict-prefetch batching threshold (batching only engages when more than 1 member is requested), so it exercises the plain, unbatched ZMSCORE lookup loop -- the baseline every wider call is compared against. Encoding: skiplist (1,000,000 elements, exceeds zset-max-listpack-entries 128). Metric of
+  interest: Ops/sec.'
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  check:
+    keyspacelen: 1
+  preload_tool:
+    run_image: redislabs/memtier_benchmark:edge
+    tool: memtier_benchmark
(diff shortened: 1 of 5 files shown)


fcostaoliveira:
Good catch — confirmed by re-reading the actual server code
(`zmscoreCommand()`, `t_zset.c`). With `ZMSCORE_PREFETCH_BATCH=16`, the
adaptive-shrink rule only splits a call into multiple `dictPrefetchKeys()`
batches once at least 32 members remain (`batch >= ZMSCORE_PREFETCH_BATCH*2`).
At 16 or 17 members the whole call is handled as a single undivided batch —
no split ever happens, so the 17-member suite didn't exercise the
straggler-avoidance shape it claimed to.

Fixed in 9822ba7:
- `16-members`: description corrected — it's a single call sized exactly at
  the batch constant, not a "batch boundary".
- `17-members`: removed.
- `32-members`: added — the true first-split boundary (even 16+16).
- `33-members`: added — 16+17, the shape a naive fixed-size chunker (no
  adaptive shrink) would instead split into 16+16+**1**, i.e. the actual
  proof the fix folds the remainder rather than stranding it.

Final family: 1 / 16 / 32 / 33 / 100.

PR #541, opened 2026-08-27
pubsub: add low-channel-count and uneven-publish-rate suites

fcostaoliveira:
The existing pubsub suites use 100 or 1000 channels, select the channel uniformly at random (`--command-key-pattern="R"`), and drive every publisher connection at the same rate. That is a good throughput benchmark, but it makes two properties invisible by construction.

### 1. Channel concentration — `mixed-10-channels-128B-publishers-170-subscribers`

Spreading publishes uniformly over 100+ channels averages away any concentration. Deployments commonly run a handful of channels with many subscribers each, which is a different regime for delivery. This suite pins 10 channels with ~17 subscribers each (170 total), keeping the 128B payload and mixed publish/subscribe structure of the existing 100-channel suites so results stay comparable across the family.

### 2. Uneven publish rate — `10-channels-128B-skewed-publishers-170-subscribers`

Where a connection is handled by one worker thread for its lifetime, worker load is determined by the *per-connection* command rate of whatever landed there. Driving every connection at the same rate therefore produces near-perfect balance regardless of the workload, and aggregate ops/sec cannot tell an evenly loaded set of workers from a badly skewed one.

Real traffic is rarely uniform — a few busy producers among many quiet ones is the common shape. This suite pairs one heavily pipelined publisher (1 connection, pipeline 30) with a pool of lightly loaded ones (48 connections, pipeline 1), over the same channels and subscribers as the suite above, so the two differ *only* in how publish rate is distributed.

With 8 workers this predicts ~3.7x max/mean worker skew against ~1.0x measured for the uniform suites — enough separation to make per-worker distribution a measurable property rather than an assumption. Pipeline depth on the hot client is the knob for dialling the ratio.

### Notes

- Both suites validate against the repo's suite tests (23 passed, 0 failed).
- The second suite needs a runner with multi-client support, since it declares three `clientconfigs`.
- The metric of interest for the second suite is the distribution of work across workers, not only aggregate ops/sec.

---

> [!NOTE]
> **Low Risk**
> Only new benchmark specification YAML; no application or infrastructure code changes.
> 
> **Overview**
> Adds **two** new memtier/pubsub-sub-bench suites under `redis_benchmarks_specification/test-suites`, both using **10 channels**, **128B** payloads, **170** subscribers (~17 per channel), **oss-standalone**, and **priority 23**.
> 
> The **mixed** suite (`memtier_benchmark-nokeys-pubsub-mixed-10-channels-128B-publishers-170-subscribers.yml`) mirrors the existing 100-channel mixed pub/sub shape but caps `--key-maximum` at 10 so channel concentration is visible instead of averaged across many channels.
> 
> The **skewed** suite (`memtier_benchmark-nokeys-pubsub-10-channels-128B-skewed-publishers-170-subscribers.yml`) keeps the same subscriber setup and drives publish with **only 3** memtier connections (`-c 3`, `--pipeline 10`) so publish work can land on a small subset of workers when connections are pinned per accept, making per-worker load distribution measurable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c70939e04f44a54740de27de1bb3dae84469aed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-nokeys-pubsub-10-channels-128B-skewed-publishers-170-subscribers.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-nokeys-pubsub-10-channels-128B-skewed-publishers-170-subscribers.yml
@@ -1,0 +1,58 @@
+version: 0.4
+name: memtier_benchmark-nokeys-pubsub-10-channels-128B-skewed-publishers-170-subscribers
+description: Publish/subscribe where publish rate is deliberately uneven across client
+  connections. Every other pubsub suite drives all publisher connections at the same rate,
+  which spreads load evenly over server/proxy worker threads and makes per-worker imbalance
+  unmeasurable by construction. Real deployments are not uniform - a small number of busy
+  producers alongside many quiet ones is the common shape, and because a connection is
+  handled by one worker for its lifetime, a single high-rate connection loads its worker far
+  harder than the rest. This suite pairs one heavily pipelined publisher with a pool of
+  lightly loaded ones so that per-worker distribution, not just aggregate throughput,
+  becomes an observable property.
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  resources:
+    requests:
+      memory: 2g
+tested-groups:
+- pubsub
+tested-commands:
+- publish
+- subscribe
+redis-topologies:
+- oss-standalone
+build-variants:
+- gcc:15.2.0-amd64-debian-bookworm-default
+- gcc:15.2.0-arm64-debian-bookworm-default
+- dockerhub
+clientconfigs:
+# Hot publisher: a single connection carrying a large share of the publishes.
+- run_image: redislabs/memtier_benchmark:edge
+  tool: memtier_benchmark
+  arguments: --test-time 120 --key-prefix "channel-" --pipeline 30 -d 128 --key-maximum
+    10 --command "PUBLISH __key__ __data__" --command-key-pattern="R" -c 1 -t 1
+    --hide-histogram
+  resources:
+    requests:
+      cpus: '2'
+      memory: 1g
+# Background publishers: many connections, each at a low per-connection rate.
+- run_image: redislabs/memtier_benchmark:edge
+  tool: memtier_benchmark
+  arguments: --test-time 120 --key-prefix "channel-" --pipeline 1 -d 128 --key-maximum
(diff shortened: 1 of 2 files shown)


fcostaoliveira:
Ran the skewed suite on real hardware (8 workers) before asking anyone to review the theory behind it. **The original design did not work, and the measurement showed why.** Pushed `c3d8b3e` with a corrected version.

### What was measured

| suite | hottest worker | max/mean |
|---|---|---|
| skewed-publishers (original design) | 14.3 % / 14.2 % | **1.14x** |
| uniform 10-channels | 12.9 % / 12.7 % | 1.03x |
| all four 100-channel suites | ~12.6 % | 1.01x |

1.14x against a 1.01–1.03x baseline is a real, reproducible effect — it came out identical on two independent builds — but it is roughly a third of what I expected, and the reason is not subtle.

### The pipelining contributed exactly nothing

The original suite opened 1 + 48 = **49 publisher connections over 8 workers**. Round-robin accept gives one worker 7 and the rest 6, and 7/49 = **14.29 %**. The measured hottest-worker shares were 14.3 % and 14.2 %.

The entire "skew" was 49 failing to divide evenly by 8. Solving for the hot connection's relative rate from the observed share gives **R = 1.0** — pipeline depth 30 bought no per-connection advantage at all, because the worker saturates rather than scaling. The suite also ran at 116k ops/sec against 240k for the otherwise-identical uniform suite, which is the same story from the throughput side.

The same arithmetic predicts the uniform sibling exactly: 200 connections / 8 workers = 25 each → 12.5 % predicted, 12.9 % and 12.7 % measured.

### What actually governs per-worker load

Not rate — **connection placement**. Opening many publisher connections puts publish work on every worker and averages the imbalance away by construction.

So the revised suite opens **3 publisher connections** and supplies rate through pipelining instead. On 8 workers that leaves 5 workers with no publish traffic at all, and predicts 33.3 % hottest share / **2.67x** max/mean.

### Sizing

Production captures of this workload shape show **2.93x–6.40x** within-node skew across 11 workers per node (cross-node imbalance is only 1.21x, so this is a single-node property and a single-node benchmark can reproduce it). 2.67x on 8 workers is the right order of magnitude rather than a token effect.

Also worth noting for anyone reviewing: in that capture, per-worker publish load correlates with connection *count* at **r = −0.098** — essentially zero. The busiest worker had fewer clients than average, and several workers with ~110 clients published nothing. Connection count is not the driver; which connections happen to be publishers is.

PR #542, opened 2026-08-31
Refresh commands.json against the Redis 8.10.1 command table

### LINKED ISSUE #527: [BUG]: commands.json is 5 minor releases behind and disagrees with the suites it validates
`commands.json` has not been refreshed in about 12 months and is now 5 minor releases behind.

```
commands.json:   396 entries, highest `since` = 8.0.0
Redis 8.10.1:    459 commands   (latest stable, released 2026-08-17)
```

Excluding `SENTINEL *` (22, arguably out of scope for a data-path benchmark spec), that leaves roughly 55 absent commands — spread across 8.2 (3), 8.4 (6), 8.6 (8), 8.8 (20) and 8.10 (17). The genuinely benchmarkable data-path subset is nearer 30; the rest are admin/cluster-internal commands where a throughput suite may not be the right instrument.

### It is not only stale, it is internally inconsistent

Five commands that suites actually run are missing from the corpus: `ARGET`, `ARRING`, `HIMPORT`, `SETX`, `CLIENT|LIST`. Because `__cli__/stats.py` looks each suite command up in `commands.json`, these are **silently dropped from coverage statistics** — so the reported coverage is computed over a corpus that does not contain everything being tested.

`SETX` is the interesting one: it is not a Redis command at all. It appears in `tested-commands` of 5 `*-expire-use-case.yml` suites and is a typo for `setex` (which 10 other suites spell correctly). A corpus check would have caught it.

`groups.json` also already knows the `array` group while `commands.json` contains no `AR*` entries, so the two files disagree.

### Two newest data types are effectively unbenchmarked

- **Vector sets** — 14 `V*` commands in the corpus, **0 suites** anywhere.
- **Array type** (new in 8.8) — 18 commands, 2 covered (`ARGET`, `ARRING`), and **no write-path suite at all** (`ARSET`, `ARINSERT`, `ARMSET`).

### Suggestion

Regenerate `commands.json` from the 8.10.1 command table, and add a scheduled CI job that diffs it against the latest stable tag and fails on drift. That prevents recurrence, and as a side effect turns "a suite names a command that does not exist" into a build failure rather than a silent statistics gap.

Happy to send the regeneration PR if that is wanted; the CI gate is probably better decided by a maintainer since it needs a release-tracking policy (latest stable, or N-1).

neoLsH:
Fixes #527

## What

Regenerates `commands.json` from the Redis **8.10.1** command table: **396 → 452 entries** (437 core + 13 vector-set + 2 carried-over RedisJSON entries), closing the 5-release gap and adding the ~56 missing commands (`MSETEX`, `INCREX`, the `AR*` auto-ring family, `BACKUP *`, `HOTKEYS *`, `HIMPORT *`, `XDELEX`, `XACKDEL`, `LMOVEM`/`BLMOVEM`, `SDIFFCARD`, `SUNIONCARD`, `VRANGE`, …). As suggested in the issue, the 22 `SENTINEL *` commands stay out of scope.

The generator is committed as `utils/regenerate_commands_json.py` so refreshes are reproducible; it supports `--check`, which exits non-zero on drift and can back a CI gate if maintainers want one (per the issue, the release-tracking policy for such a job is a maintainer decision, so no workflow change here).

## How the corpus was rebuilt

- Entry names derive from `src/commands/*.json` file names, preserving subcommand hyphens (`CLIENT NO-EVICT`, `CLUSTER SET-CONFIG-EPOCH`) to keep the existing corpus spelling.
- `acl_categories` replicate `setImplicitACLCategories()` from `src/server.c` (explicit categories plus derived `@read`/`@write`/`@fast`/`@slow`/…), in the canonical `src/acl.c` order. Cross-checked against `COMMAND INFO` on a live 8.10.1 instance: **zero divergence across all 449 reported commands**.
- The previously missing `HIMPORT`, `ARGET`, `ARRING` suite commands now resolve, and `groups.json`'s `array` group agrees with the corpus again.
- `JSON.*` entries not present in the checkout are carried over verbatim.
- Note: `arguments` no longer carry `display_text` — the source command table does not publish it anymore.

Second commit fixes the `setx` typo the issue calls out: it is dropped from `tested-commands` of the 5 `*-expire-use-case.yml` suites (they already list the correct `setex`).

## Verification

- `redis-benchmarks-spec-cli --tool stats --fail-on-required-diff` → exit 0
- `python3 utils/regenerate_commands_json.py <redis 8.10.1 checkout> --check` → `commands.json matches the checkout`
- Field-level diff against the old corpus audited: remaining deltas are genuine 8.0 → 8.10.1 evolution (e.g. new `sentinel` availability flags, `notes`/`incomplete` key-spec enrichments)

---

> [!NOTE]
> **Low Risk**
> Changes are corpus metadata and test-suite command lists for the stats CLI, not runtime Redis behavior; main risk is incorrect command metadata affecting reporting or validation.
> 
> **Overview**
> **Regenerates `commands.json` from the Redis 8.10.1 command table**, growing the corpus from ~396 to **452** entries (core commands, vector-set module, plus carried-over `JSON.GET`/`JSON.SET`) and adding missing commands such as `MSETEX`, `INCREX`, auto-ring `AR*`, `BACKUP *`, `HOTKEYS *`, `HIMPORT *`, stream helpers, and related list/set commands. **`SENTINEL *` stays excluded** as out of scope for this benchmark corpus.
> 
> Adds **`utils/regenerate_commands_json.py`** so refreshes are reproducible from a Redis checkout, with **`--check`** to fail on drift against the committed file. Regenerated metadata follows upstream naming (including hyphenated subcommands), **replicates implicit ACL categories** like `setImplicitACLCategories()`, and **drops `display_text` on arguments** where the upstream table no longer provides it.
> 
> Fixes benchmark suite hygiene by **removing the erroneous `setx` entry** from the `*-expire-use-case` suites (they already list `setex`), so stats validation treats unknown tested commands as errors rather than silent gaps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eab3e8e3436698906d5df9f368ed3f7638ce6a9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -7,7 +7,10 @@
     "acl_categories": [
       "@slow"
     ],
-    "arity": -2
+    "arity": -2,
+    "command_flags": [
+      "sentinel"
+    ]
   },
   "ACL CAT": {
     "summary": "Lists the ACL categories, or the commands inside a category.",
@@ -22,14 +25,14 @@
       {
         "name": "category",
         "type": "string",
-        "display_text": "category",
         "optional": true
       }
     ],
     "command_flags": [
       "noscript",
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
     ]
   },
   "ACL DELUSER": {
@@ -47,7 +50,6 @@
       {
         "name": "username",
         "type": "string",
-        "display_text": "username",
         "multiple": true
       }
     ],
@@ -55,7 +57,8 @@
       "admin",
       "noscript",
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
     ],
     "hints": [
       "request_policy:all_nodes",
@@ -76,18 +79,15 @@
     "arguments": [
       {
         "name": "username",
-        "type": "string",
-        "display_text": "username"
+        "type": "string"
       },
       {
         "name": "command",
-        "type": "string",
-        "display_text": "command"
+        "type": "string"
       },
       {
         "name": "arg",
         "type": "string",
-        "display_text": "arg",
         "optional": true,
         "multiple": true
       }
@@ -96,8 +96,10 @@
       "admin",
       "noscript",
       "loading",
-      "stale"
-    ]
+      "stale",
+      "sentinel"
+    ],
+    "history": []
   },
   "ACL GENPASS": {
     "summary": "Generates a pseudorandom, secure password that can be used to identify ACL users.",
@@ -112,14 +114,14 @@
       {
         "name": "bits",
         "type": "integer",
-        "display_text": "bits",
         "optional": true
       }
     ],
     "command_flags": [
       "noscript",
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
     ]
   },
   "ACL GETUSER": {
@@ -127,16 +129,6 @@
     "since": "6.0.0",
     "group": "server",
     "complexity": "O(N). Where N is the number of password, command and pattern rules that the user has.",
-    "history": [
-      [
-        "6.2.0",
-        "Added Pub/Sub channel patterns."
-      ],
-      [
-        "7.0.0",
-        "Added selectors and changed the format of key and channel patterns from a list to their rule representation."
-      ]
-    ],
     "acl_categories": [
       "@admin",
       "@slow",
@@ -146,15 +138,25 @@
     "arguments": [
       {
         "name": "username",
-        "type": "string",
-        "display_text": "username"
+        "type": "string"
       }
     ],
     "command_flags": [
       "admin",
       "noscript",
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
+    ],
+    "history": [
+      [
+        "6.2.0",
+        "Added Pub/Sub channel patterns."
+      ],
+      [
+        "7.0.0",
+        "Added selectors and changed the format of key and channel patterns from a list to their rule representation."
+      ]
     ]
   },
   "ACL HELP": {
@@ -168,7 +170,8 @@
     "arity": 2,
     "command_flags": [
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
     ]
   },
   "ACL LIST": {
@@ -186,7 +189,8 @@
       "admin",
       "noscript",
       "loading",
-      "stale"
+      "stale",
+      "sentinel"
     ]
   },
   "ACL LOAD": {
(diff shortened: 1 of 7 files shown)


neoLsH:
Thanks — both merge asks plus the typo-class gate are in eab3e8e.

**Carry-over made explicit.** `CARRIED_OVER_COMMANDS = ("JSON.GET", "JSON.SET")` replaces the `existing - regenerated` computation, so `--check` no longer validates against the very file it checks. Anything present only in the committed corpus now fails the run outright (`commands present only in the committed corpus (removed upstream?): [...]`) — verified by planting a ghost entry. That also covers your parenthetical: no `SENTINEL*` can ride along, and a future upstream removal of either JSON command is a build failure, not a silent re-emit.

**Name derivation moved off the filename.** You were right that the 8.10.1 entries carry `container` (169 of 459 files), so the name is now `container + " " + key`, with standalone commands keeping their key verbatim (`MSETEX`, `RESTORE-ASKING` — whose key is itself hyphenated upstream). That deletes the first-hyphen rule and the `NON_SUBCOMMAND_HYPHENATED` escape hatch; sentinel exclusion is now `stem == "sentinel" or stem.startswith("sentinel-")`. The regenerated output is byte-identical to what was already committed, i.e. the heuristic happened to agree with the authoritative spelling on all 452 entries — but now it's by construction.

**tested-commands resolution check: yes, here rather than follow-up.** The script now validates every suite's `tested-commands` against the regenerated corpus (as-is and upper-cased, like `stats.py`, plus pipe→space so `client|list` resolves against `CLIENT LIST`). Re-introducing `setx` locally fails the run with the offending file and entry named. It closes exactly the class you traced: entries seeded into `tested_commands` before the derivation loop are never looked up, so bogus names sailed through. If maintainers later wire the script into CI, this gate comes for free with `--check`.

**`--check` brittleness.** Now compares parsed JSON, so a trailing newline (or any whitespace-only formatting difference) no longer reports drift; the written file keeps the historical byte shape (no final newline) to avoid churn in the committed corpus. Verified by appending a newline and re-running `--check` — still green.

On CI: agreed `Validate SPEC fields`/`tox` need to run before merge — they're pending workflow approval for the fork head, and the local `--fail-on-required-diff` exit 0 stands as the interim evidence.

Issue #548, opened 2026-09-02
Add ZRANGEBYLEX / ZREVRANGEBYLEX / ZLEXCOUNT / ZRANGESTORE BYLEX coverage


fcostaoliveira:
## Gap

The `test-suites/` directory has **zero** benchmark specs exercising the lexicographical range path of the sorted-set commands:

- `ZRANGEBYLEX` — no spec
- `ZREVRANGEBYLEX` — no spec
- `ZLEXCOUNT` — no spec
- `ZRANGESTORE key src min max BYLEX` — no spec
- `ZRANGE key min max BYLEX [REV]` — no spec

Verified with:

```
grep -A1 "tested-commands:" test-suites/*.yml | grep -iE "bylex|zlex"
# → no matches
```

For context, the ZRANGEBYSCORE side is covered by 2 specs today:

```
memtier_benchmark-1key-zset-100-elements-zrangebyscore-all-elements.yml
memtier_benchmark-1key-zset-100-elements-zrangebyscore-all-elements-long-scores.yml
```

## Why it matters

`t_zset.c` has a distinct lexicographical code path (`zslParseLexRange`, `zslNextValidElement`, `zslIsInLexRange`, `sdscmplen`-based comparisons instead of double compares, min/max sentinels `-`/`+`/`(`/`[`). This whole branch is currently invisible to every optimization/regression sweep — anything that touches skiplist iteration, encoding conversion, or (recently) B-tree POCs can regress the by-lex path without any test in the suite noticing.

Recent same-runner comparisons on the zset B-tree POC family (bulkappend / bulk-dict-insert / same-score) touch the sorted-set iteration paths heavily; the by-score half is validated but the by-lex half isn't, which is a real coverage blind spot for any zset-line change proposed for upstream.

## Starter set of proposed specs

Following the pattern of the existing by-score specs (1-key, 100 elements as the baseline size), and mirroring shapes that already exist for the by-score variants elsewhere in the suite:

| Spec (proposed) | Command | Element count | Notes |
|---|---|---|---|
| `1key-zset-100-elements-zrangebylex-all-elements` | ZRANGEBYLEX | 100 | mirror of by-score baseline |
| `1key-zset-1M-elements-zrangebylex-100-elements-pipeline-10` | ZRANGEBYLEX | 1M zset, 100-element slice | large-key iteration |
| `1key-zset-100-elements-zrevrangebylex-all-elements` | ZREVRANGEBYLEX | 100 | reverse walk |
[...]

PR #554, opened 2026-09-06
Add by-lex sorted-set coverage: ZRANGEBYLEX / ZREVRANGEBYLEX / ZLEXCOUNT / ZRANGESTORE BYLEX

neoLsH:
Implements the five-spec starter set proposed in #548 — the lexicographical half of the sorted-set range family (`zslParseLexRange` / `zslNextValidElement` / `sdscmplen` comparisons, `-`/`+` sentinels) had zero coverage, so any zset-line change (skiplist iteration, encoding conversion, B-tree POCs) could regress the by-lex path unnoticed.

## What's added

| Spec | Command | Shape | Mirrors |
|---|---|---|---|
| `1key-zset-100-elements-zrangebylex-all-elements` | ZRANGEBYLEX | 100-element listpack, full `- +` scan | `zrangebyscore-all-elements` baseline |
| `1key-zset-100-elements-zrevrangebylex-all-elements` | ZREVRANGEBYLEX | 100-element listpack, reverse walk (`+ -`, max first) | reverse sibling of the above |
| `1key-zset-1M-elements-zrangebylex-100-elements-pipeline-10` | ZRANGEBYLEX | 1M skiplist, exact 100-element slice via `LIMIT 0 100`, pipeline 10 | large-key iteration |
| `1key-zset-1M-elements-zlexcount-pipeline-10` | ZLEXCOUNT | 1M skiplist, full-domain count-only, pipeline 10 | pairs with the slice spec to isolate reply-materialisation cost |
| `1key-zset-600K-elements-zrangestore-300K-elements-bylex` | ZRANGESTORE | 600K → exact 300K copy via `- + BYLEX LIMIT 0 300000` | by-lex mirror of the by-score `zrangestore-300K` hero test |

All in `oss-standalone`, `sorted-set` group, same build variants / priorities as their by-score counterparts.

## Design notes

**Constant-score datasets.** BYLEX semantics are only meaningful when all members share one score (with distinct scores, Redis filters in `(score,member)` order and the "lexicographical" result order is an artifact). So:
- the two 1M specs **reuse the existing `1key-zset-1M-elements-const-score` dataset** (from the zmscore family — single-`__key__` preload, verified exact 1,000,000, no #508 dual-draw shortfall);
- the 100-elements pair adds `1key-zset-100-elements-const-score` (same member list as the by-score `100-elements-float` dataset, constant score 0);
- the ZRANGESTORE spec adds `1key-zset-600K-elements-const-score` (members 1..600001, constant score 0 — exact cardinality for the same single-draw reason).

**Exact slices via LIMIT.** The proposed `600K → 300K` copy cannot be expressed as a lex *value* range: unpadded integer members have no lexicographic halfway point (`"3..."` covers far more than half the domain). `LIMIT` — valid precisely with `BYSCORE`/`BYLEX` — makes every slice exact: `LIMIT 0 100` for the 1M read spec, `LIMIT 0 300000` for the store spec.

**No dependency on #542**: all four commands are already in the current `commands.json`.

## Verification

- Live smoke against a local unstable build (`redis-server v=8.9.241 sha=80920a7d`): for each spec, parsed the yml and executed its `init_commands`/preload semantics + client command — keyspacelen=1, listpack encoding at 100 elements, both full scans return all 100 members in ascending/descending lex order, `LIMIT 0 100` returns exactly 100, `ZRANGESTORE ... BYLEX LIMIT 0 300000` copies exactly 300K elements forming a lex prefix of the source, `ZLEXCOUNT - +` equals `ZCARD`. 14/14 checks pass.
- `redis-benchmarks-spec-cli --tool stats --fail-on-required-diff` exits 0; ERROR count identical to the main baseline (pre-existing `array-*` spec noise, untouched).

---

> [!NOTE]
> **Low Risk**
> Adds benchmark specification YAML only; no Redis server or client code changes.
> 
> **Overview**
> Adds **five new memtier_benchmark YAML specs** so lexicographical sorted-set commands (`ZRANGEBYLEX`, `ZREVRANGEBYLEX`, `ZLEXCOUNT`, `ZRANGESTORE ... BYLEX`) are exercised in CI—parallel to existing by-score coverage, which previously had no BYLEX counterpart.
> 
> The **100-member pair** uses inline `ZADD` init with **constant score 0** (`1key-zset-100-elements-const-score`) and full `- +` / `+ -` scans on **listpack** encoding. **1M-member specs** reuse **`1key-zset-1M-elements-const-score`**: one runs **`ZLEXCOUNT lb - +`** with pipeline 10 (skiplist rank path), the other **`ZRANGEBYLEX`** with a **random lex lower bound** and `LIMIT 0 100` so each call does a real seek instead of always hitting the skiplist head. The **600K spec** preloads **`1key-zset-600K-elements-const-score`** and benchmarks **`ZRANGESTORE zset1 zset - + BYLEX LIMIT 0 300000`** as the by-lex mirror of the existing by-score `zrangestore` hero test.
> 
> All specs target **`oss-standalone`**, **`sorted-set`**, and the same GCC/dockerhub build variants and priorities as their by-score siblings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d689914c1bad68fb52861ad58ee9206b0800848. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


neoLsH:
Both points confirmed against `t_zset.c` and fixed in 7d68991 — thanks, the ZLEXCOUNT one was a genuine factual error in my description.

**ZLEXCOUNT does not traverse at skiplist encoding.** Verified: `zlexcountCommand`'s SKIPLIST branch resolves both endpoints with `zslNthInLexRange` (`nth=0` then `nth=-1`, each yielding its rank) and derives the count by rank arithmetic — `count = length-(rank_first-1)`, then `-= (length-rank_last)` — which is the `O(log(N))` in `commands.json`. Only the LISTPACK branch has the `while (eptr) { ... zzlNext }` loop, and this spec is 1M/skiplist. Dropped the "walks the same iteration path" claim and the "isolates reply-materialisation from pure lex traversal" framing (the pairing doesn't hold: one is `O(log N)` rank-only, the slice spec is `O(log N + M)` seek+traverse). The description now states what it actually covers — `zslParseLexRange` on the `-`/`+` sentinels, both `*InLexRange` seeks, and the `zslGetRank` descents — and notes that at this command cost with `-c 1 -t 1 --pipeline 10` much of the recorded Ops/sec is dispatch/protocol overhead rather than server-side lex work.

**Adopted the random lex start.** The 1M ZRANGEBYLEX spec now uses `[__key__ + LIMIT 0 100` (`--command-key-pattern=R --key-minimum 1 --key-maximum 1000000 --key-prefix ""` so no `memtier-` prefix corrupts the integer lex domain), renamed to `...zrangebylex-random-position-pipeline-10` to mirror the existing `zrangebyscore-random-position`. Confirmed the degeneracy you described: with the `-` min sentinel, `zslLexValueGteMin` is unconditionally true so `zslFirstInLexRange` lands on the leftmost node — live, `- + LIMIT 0 100` returns `1,10,100,...` on every call. With the random start, `[50000 + LIMIT 0 100` seeks to `50000` and returns a full 100-slice, and near-tail starts (`[99999`) return short slices as expected for a uniform draw — documented in the description rather than left as a surprise.

Re-ran the local smoke (now 19 checks incl. the random-position seek, the `- +` degeneracy contrast, and a `ZLEXCOUNT [50000 +` vs `ZRANGEBYLEX [50000 +` window-cardinality match) — all pass; `--tool stats --fail-on-required-diff` exits 0 with the same pre-existing `array-*` ERROR count as main.

Issue #496, opened 2026-08-15
Add benchmark coverage for glob pattern matching (KEYS, SCAN MATCH, PSUBSCRIBE)


fcostaoliveira:
Glob pattern matching runs on several hot paths, but the suite barely touches it.

`stringmatchlen()` in `util.c` is the shared matcher behind `KEYS`, the `MATCH` option on `SCAN`/`HSCAN`/`SSCAN`/`ZSCAN`, pattern pubsub delivery, and ACL key-pattern checks. It is a recursive matcher, so its cost scales with pattern shape, not just with the number of keys visited. That makes it easy to regress without noticing.

Current state across the 432 registered specs:

- 1 spec passes a `MATCH` pattern: `memtier_benchmark-1key-zset-1K-elements-zscan-match-count-100`.
- 0 specs run `KEYS` with a pattern.
- 0 specs run `PSUBSCRIBE`. The five pubsub specs all `PUBLISH` to literal channel names, so they never reach the pattern-matching path.

There is a subtlety that makes the gap wider than the counts suggest. Both call sites short-circuit on the bare `*` pattern before they ever call the matcher:

- `db.c`: `allkeys = (pattern[0] == "*" && plen == 1);` and the loop then tests `if (allkeys || stringmatchlen(...))`.
- `db.c`: `use_pattern = !(patlen == 1 && pat[0] == "*");` guards every `SCAN`-family call.

So a `KEYS *` or a `SCAN ... MATCH *` spec would add no coverage at all. Only a non-trivial pattern exercises the matcher.

Suggested additions:

- `KEYS` against a large keyspace with a selective prefix pattern, plus a pattern with a leading wildcard, since the two produce very different recursion behavior.
- `SCAN ... MATCH <prefix>*` with a realistic `COUNT`, so per-element matching dominates rather than the cursor walk.
- `HSCAN` and `SSCAN` with `MATCH`, to cover the listpack and hashtable encodings separately.
- A pattern pubsub spec: subscribers on `PSUBSCRIBE news.*` style patterns with publishers on matching channels, which measures the per-pattern loop on the publish path.

Happy to help shape the workloads if useful.

PR #556, opened 2026-09-07
Add glob-pattern matcher coverage: KEYS, SCAN MATCH, HSCAN and SSCAN

neoLsH:
Addresses #496 — adds benchmark coverage for the glob-pattern matcher (`stringmatchlen()` in `util.c`), which the suite barely touched.

## What this adds

Seven specs, all driving the matcher through `memtier_benchmark`:

| Spec | Command / pattern | Encoding | Why |
|---|---|---|---|
| `1Mkeys-generic-keys-selective-prefix-match` | `KEYS memtier-12345*` | — | prefix-anchored pattern → low-backtracking path |
| `1Mkeys-generic-keys-leading-wildcard-match` | `KEYS *2345` | — | leading `*` → suffix-match-at-every-position path |
| `1Mkeys-generic-scan-match-count-100` | `SCAN 0 MATCH memtier-12345* COUNT 100` | — | per-element matching dominates the cursor walk |
| `1key-hash-hscan-1K-fields-match-count-100` | `HSCAN test_hset 0 MATCH field:1* COUNT 100` | hashtable (1000 fields) | dict scan path |
| `1key-hash-hscan-50-fields-match-count-100` | `HSCAN test_hset50 0 MATCH field:1* COUNT 100` | listpack (50 fields) | listpack scan path |
| `1key-set-1K-elements-sscan-match-count-100` | `SSCAN test_sadd 0 MATCH member:1* COUNT 100` | hashtable (1000 members) | dict scan path |
| `1key-set-100-elements-sscan-match-count-100` | `SSCAN test_sadd100 0 MATCH member:1* COUNT 100` | listpack (100 members) | listpack scan path |

HSCAN and SSCAN are each split across the two encodings, as suggested, so the hashtable and listpack iteration paths are priced separately.

## On the bare-`*` subtlety

Every pattern here is non-trivial on purpose. As the issue notes, both call sites short-circuit before the matcher — `db.c` sets `allkeys = (pattern[0] == "*" && plen == 1)` for KEYS and `use_pattern = !(patlen == 1 && pat[0] == "*")` for the SCAN family — so a `KEYS *` or `SCAN ... MATCH *` spec would add zero matcher coverage. The prefix vs. leading-wildcard KEYS pair is there specifically to separate the two recursion behaviours of `stringmatchlen()`. Match counts were verified locally against a preloaded keyspace (`memtier-12345*` → 11/1,000,000; `*2345` → 100/1,000,000; `field:1*` → 112/1000 hashtable, 11/50 listpack; `member:1*` → 112/1000 hashtable, 12/100 listpack).

`tested-commands` / `tested-groups` line up with `commands.json` (KEYS/SCAN → generic, HSCAN → hash, SSCAN → set).

## The pattern-pubsub item — deferred, and a question

The fourth suggestion (subscribers on `PSUBSCRIBE news.*` with publishers on matching channels) is **not** in this PR, and I wanted to flag why rather than quietly drop it.

The existing pubsub specs drive the subscriber side with `pubsub-sub-bench`, and as far as I can tell from its documented flags it only exposes channel-based subscribe (`-channel-minimum` / `-channel-maximum` / `-subscriber-prefix` / `-subscribers-per-channel`) — there's no PSUBSCRIBE / pattern mode. So a pattern-pubsub workload that actually reaches the per-pattern loop on the publish path can't be driven with the current subscriber tooling.

@fcostaoliveira since `pubsub-sub-bench` is your tool: would you want a `-mode psubscribe` (or a `-subscriber-pattern`) added there first, after which I'm happy to contribute the matching spec? Or if there's already a way to drive pattern subscriptions that I've missed, point me at it and I'll fold a pubsub spec into this PR. I kept the matcher-path specs separate so they're not blocked on that.

---

> [!NOTE]
> **Low Risk**
> Adds benchmark YAML definitions only; no runtime or Redis server code changes.
> 
> **Overview**
> Adds **seven new memtier benchmark specs** (YAML only) to exercise Redis’s shared glob matcher (`stringmatchlen`) on paths that existing suites mostly skipped because bare `*` bypasses matching.
> 
> **KEYS** on a 1M-key space uses two deliberate patterns: prefix `memtier-12345*` (low-backtracking) and leading-wildcard `*2345` (backtracking at every position). **SCAN** gets the same selective prefix with `MATCH` and `COUNT 100`, paralleling the existing ZSCAN MATCH spec.
> 
> **HSCAN** and **SSCAN** each add **hashtable vs listpack** pairs with structured field/member names and `field:1*` / `member:1*` patterns so matcher cost is measured on dict and listpack iteration separately. Pattern pubsub coverage is explicitly out of scope here.
> 
> All specs use the usual preload + single-client `memtier_benchmark` workload shape, oss-standalone, and the standard gcc/dockerhub build matrix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11254622092a497f653c08f99e02d9e33c76f0b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-generic-keys-leading-wildcard-match.yml
+++ b/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-generic-keys-leading-wildcard-match.yml
@@ -1,0 +1,43 @@
+version: 0.4
+name: memtier_benchmark-1Mkeys-generic-keys-leading-wildcard-match
+description: Runs KEYS with a leading-wildcard glob pattern against the same 1M-key keyspace as the selective-prefix
+  variant. A pattern that begins with `*` forces stringmatchlen() to attempt the suffix match at every
+  position of every key -- the recursive backtracking path -- so the per-key matcher cost differs sharply
+  from a prefix-anchored pattern even though both walk all 1M keys. `*2345` matches 100 of 1,000,000 keys
+  (verified locally). Together the two KEYS specs price the two recursion behaviours of the shared matcher
+  that KEYS, SCAN MATCH, HSCAN, SSCAN and pattern pubsub all depend on.
+dbconfig:
+  configuration-parameters:
+    save: '""'
+  check:
+    keyspacelen: 1000000
+  preload_tool:
+    run_image: redislabs/memtier_benchmark:edge
+    tool: memtier_benchmark
+    arguments: '"--data-size" "10" "--ratio" "1:0" "--key-pattern" "P:P" "-c" "50" "-t" "2" "--hide-histogram"
+      "--key-minimum" "1" "--key-maximum" "1000000" "-n" "allkeys" --pipeline 50'
+  resources:
+    requests:
+      memory: 1g
+  dataset_name: 1Mkeys-string-10B-size
+  dataset_description: This dataset contains 1 million string keys, each with a data size of 10 bytes.
+tested-commands:
+- keys
+tested-groups:
+- generic
+redis-topologies:
+- oss-standalone
+build-variants:
+- gcc:15.2.0-amd64-debian-bookworm-default
+- gcc:15.2.0-arm64-debian-bookworm-default
+- dockerhub
+clientconfig:
+  run_image: redislabs/memtier_benchmark:edge
+  tool: memtier_benchmark
+  arguments: --command "KEYS *2345" --command-key-pattern R --hide-histogram --test-time 180 --pipeline
+    1 -c 1 -t 1
+  resources:
+    requests:
+      cpus: '2'
+      memory: 2g
+priority: 44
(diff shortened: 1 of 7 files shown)


neoLsH:
Thanks — the off-by-one was real. Fixed in 1125462: bumped the four object preloads to `--key-maximum` N+1 (1001 / 51 / 1001 / 101), matching the zscan / restore / mset siblings. You're right that it fails silently, since `check.keyspacelen` counts DB keys, not fields/members.

I re-verified the match figures against a local redis-server — so through the real `stringmatchlen`, not just arithmetic. With the corrected 1000 / 50 / 1000 / 100 datasets: `field:1*` → 112 and 11, `member:1*` → 112 and 12, which is what the descriptions claim. Left the 1M trio alone as you suggested; its preload is verbatim from `1Mkeys-string-get-10B`, and 11 / 100 don't move between 999,999 and 1,000,000.

On `util.c` in `files-to-groups.json` — I'd lean toward a separate change, and I'm not sure a narrow mapping is even the right lever. `util.c` is genuinely broad; `stringmatchlen` is one of many helpers in there, and mapping the file to `generic`/`hash`/`set`/`sorted-set` would narrow what currently (and safely) falls back to `core-specs.json` for every *other* `util.c` change. If the goal is to make these specs fire on a `stringmatchlen` change, adding them to `core-specs.json` might be cleaner than pinning `util.c` to four groups. Either way it's a CI-scoping call that outlives this PR, so happy to open a follow-up rather than fold it in here.

On the listpack `COUNT` — agreed it's a no-op there. I kept `COUNT 100` identical across the hashtable and listpack variants on purpose, so the encoding is the only variable between them, but you're right that `-match-count-100` in the two listpack filenames names a knob that isn't doing anything. I can drop it from those two names if you'd prefer the filename to reflect what actually runs.

Issue #509, opened 2026-08-17
Shared 10K-element list preload (RPUSH strlist, S key-pattern) actually populates 9,999 elements


fcostaoliveira:
## Summary

The shared `1key-list-10K-elements-string`/`-integer` preload (`RPUSH strlist __key__
--key-maximum 10000 --key-minimum 1 -n allkeys --key-prefix "hello" --command-key-pattern S`)
actually populates **9,999** elements, not 10,000 as claimed in every sibling spec's
`dataset_description`.

## Evidence

Reproduced locally against a real `redis-server` + `redislabs/memtier_benchmark:edge` (the
exact preload command copied verbatim from `memtier_benchmark-1key-list-10K-elements-lindex-string.yml`):

```
--hide-histogram --command "RPUSH strlist __key__" --key-maximum 10000 \
  --key-minimum 1 -n allkeys --key-prefix "hello" --command-key-pattern S -c 1 -t 1 --pipeline 50
```

Result: `LLEN strlist` = **9999**. First element `hello1`, last element `hello9999` (never
`hello10000`).

## Affected specs (all claim "10,000 string/integer elements" in `dataset_description`)

- `memtier_benchmark-1key-list-10K-elements-lindex-string.yml`
- `memtier_benchmark-1key-list-10K-elements-lindex-string-pipeline-10.yml`
- `memtier_benchmark-1key-list-10K-elements-lpos-string.yml`
- `memtier_benchmark-1key-list-10K-elements-linsert-lrem-integer.yml`
- `memtier_benchmark-1key-list-10K-elements-linsert-lrem-string.yml`
- `memtier_benchmark-1key-list-10K-elements-lindex-integer.yml`
- `memtier_benchmark-1key-list-10K-elements-lpos-integer.yml`

(Found while reviewing #441, which reuses this same preload for its new LSET-on-quicklist spec
and correctly documents the true count as 9,999 -- flagged during review as contradicting the
sibling specs, then verified empirically as above. #441's own spec is unaffected in *purpose*
since LSET index 5000 is valid either way; only the pre-existing specs' `dataset_description`
text is inaccurate.)

## Suggested fix

Either bump `--key-maximum` to 10001 (S/sequential key-pattern with `-n allkeys` appears to stop
one short of `key-maximum`, same off-by-one class as the P-pattern issue in #508) to get a true
10,000-element list, or correct the `dataset_description` text in all 7 files to say 9,999.
[...]

PR #558, opened 2026-09-10
test-suites: add a saturated (-c 50) BITOP XOR short-tail spec

fcostaoliveira:
## Why

`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10` was added to scope a reported
BITOP short-tail regression, and it has the right command, the right value size and the
right pipeline depth — but it runs at `-c 1 -t 1`. With a single connection, pipeline 10
leaves the server idle between batches, so throughput is set by round-trip time rather
than by the cost of the command. A few-percent change in the per-command path is diluted
below run-to-run noise, and the spec returns a confident flat number.

That is not hypothetical. Across the `391e3452` → `8dfb823c` boundary, interleaved 3 trials
per side on a dev box (not the fleet), same binaries and same pipeline depth, changing only
the connection count:

| connections | Ops/sec, suspect vs parent | trial ranges |
| --- | ---: | --- |
| `-c 1` | ~2% apart | overlapping |
| `-c 50` | ~4–5% apart | non-overlapping |

Absolute throughput roughly doubles from `-c 1` to `-c 50`, which is the tell that the
single-connection run was nowhere near saturating the main loop.

## What this adds

`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10-c50` — identical to the existing
spec except for the load model:

- 50 client connections (`-c 25 -t 2`), matching the default `redis-benchmark` uses, spread
  over 2 threads so the client is not the bottleneck
- same 10 × 50-byte RAW source keys, so an 18-byte scalar tail remains after the first AVX2
  block and the short-tail path shared by AND/OR/XOR and the newer DIFF/DIFF1/ANDOR/ONE
  operations is still what's exercised
- same `--pipeline 10`, same `--test-time 180`

Metric of interest is Ops/sec.

I've left the `-c 1` spec in place rather than editing it, so existing history stays
comparable — but for a BITOP command-cost question this saturated variant is the one to
read.

## Notes

- Coordinators load specs from the installed pip release, so this needs a version bump and a
  reinstall on the runners before it will run.
- Scopes redis/redis#15289.

---

> [!NOTE]
> **Low Risk**
> Adds benchmark YAML only; no runtime or Redis code changes.
> 
> **Overview**
> Adds a new memtier benchmark test suite **`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10-50_conns`** so BITOP XOR short-tail performance can be measured under saturated load.
> 
> It mirrors **`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10`** (same 10 × 50-byte keys, `BITOP XOR`, pipeline 10, 180s run) but drives the server with **50 connections** (`-c 25 -t 2`) instead of `-c 1 -t 1`, and bumps client CPU to **3**. The intent is Ops/sec that reflects **per-command BITOP cost** rather than RTT-limited throughput on a single connection, for scoping **redis/redis#15289**. The existing `-c 1` spec is left unchanged for historical comparability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2dadd1079f6776b8ba4942432c1b9938794ca98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Thanks — both of the two blocking points were worth raising. Addressed in d2dadd1.

**Filename convention: you're right, fixed.** Renamed to
`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10-50_conns`. I checked the tree
before changing it: of 509 specs, zero use a bare `-cNN` suffix, and the `_conns` family
consistently encodes the *total* (`-400_conns` is `-c 40 -t 10`, `-2000_conns` is
`-c 200 -t 10`). This spec is `-c 25 -t 2`, so 50 total connections. `-c50` was doubly
misleading since `-c` here is 25.

**Preload landing 10 keys: verified, no change needed.** The premise that neither spec has
run on the fleet isn't right — the `-c 1` sibling, whose preload block this copies verbatim,
has run extensively. There are 389 timeseries under
`memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10` spanning both arches, several
platforms and multiple orgs, with real values (e.g. 258469.54 Ops/sec on
`x86-aws-m7i.metal-24xl` at 8.9.241). Since `check: keyspacelen: 10` is enforced at benchmark
start and would fail the run on a 9-key preload, those passing runs are positive evidence that
the sequential preload lands exactly 10 keys.

Worth noting *why* it's fine, since the #509 comparison was apt: that bug was in the request
*count* derived for `-n allkeys`, which stopped one short of the range. Here `-n 10` is
explicit, so nothing derives the count and the generator just walks `key-minimum..key-maximum`
inclusive.

**On the over-claim — fair, reworded.** You're right that doubling from `-c 1` establishes
`-c 1` was under-saturated, not that `-c 50` is at saturation. The description now says the
measured Ops/sec "reflects the per-command cost of BITOP rather than client-server round-trip
time" instead of asserting the main loop is CPU-saturated. I'd rather not put a server-side CPU
number in the description when I haven't measured it on the fleet host this spec will actually
run on.

PR #562, opened 2026-09-11
test-suites: add a saturated 11-hot-key MSET overwrite spec

fcostaoliveira:
Adds `memtier_benchmark-11keys-mset-overwrite-11-pairs-12B-pipeline-10-50_conns`.

### What it runs

`MSET` with 11 key/value pairs per command, `--pipeline 10` over 50 client connections
(`-c 25 -t 2`), repeatedly overwriting a hot keyspace of 11 short-named STRING keys holding
12-byte EMBSTR values. Metric of interest is **p50 latency** (lower is better).

### Why, when a 100-key MSET overwrite spec already exists

`memtier_benchmark-100keys-mset-overwrite-10B-pipeline-1` covers the same subsystem, but it is
`-c 1 -t 1 --pipeline 1` with 3 pairs per command, and it cannot answer a per-command-cost
question. Two independent reasons, both of which I hit in practice:

1. **Concurrency.** At one connection and pipeline 1 the server is idle between requests, so
   round-trip time sets the throughput and a few-percent change in the command path is diluted
   below run-to-run noise.
2. **Metric resolution.** At that shape p50 comes back at 0.031 ms. memtier reports three
   decimals, so the next representable value is 0.032 — roughly a 3.2% step. An effect smaller
   than one step cannot be expressed at all, and "p50 identical on both sides" is then
   quantization, not evidence of no change. At pipeline 10 over 50 connections p50 sits near
   1.0 ms, where 0.001 ms is about 0.1%.

I ran the `-c 1` spec against a reported MSET p50 regression and read its flat result as "no
regression." That was the wrong conclusion from a measurement that could not have shown one.

### Key length is deliberate

This scopes redis/redis#15320, which concerns a `dictFind` **key-length** fast path, so key-name
lengths are matched to the report rather than normalized: `test_MSET1..test_MSET11` gives nine
10-byte names and two 11-byte names, against the report's ten 10-byte and one 11-byte. Using a
uniform-length prefix here would risk tuning out the very path under test.

The 11-key keyspace keeps the main dict and every key object resident in L1/L2 with no
rehashing and effectively no collisions, so `dictFind` is cache-latency dominated — unlike the
1M–10M distinct-key MSET load specs, which are DRAM-latency dominated and cannot show a
key-comparison difference.

`cpus: '3'` for `-t 2` keeps the client clear of the saturation trap seen in #485/#486/#487.

The `-c 1` sibling is left in place for history continuity.

### Not yet run on the fleet

This spec has not run on the fleet yet — the preload block follows the same
`--key-prefix`/`--key-minimum`/`--key-maximum` pattern as the merged 100-key sibling, and
`check: keyspacelen: 11` will fail the run rather than report a wrong number if it lands short.

---

> [!NOTE]
> **Low Risk**
> Benchmark-only YAML addition with no runtime or product code changes; misconfiguration would fail checks rather than affect production.
> 
> **Overview**
> Adds a new memtier benchmark suite **`memtier_benchmark-11keys-mset-overwrite-11-pairs-12B-pipeline-10-50_conns`** aimed at measuring **MSET** per-command cost via **p50 latency** under load, not the existing low-concurrency 100-key overwrite spec.
> 
> The workload **preloads 11 STRING keys** (`test_MSET1..11`, 12-byte EMBSTR values) and then runs **11 pair MSET** with **`--command-key-pattern=S`**, **`--pipeline 10`**, and **50 connections** (`-c 5 -t 10`), repeatedly overwriting that tiny hot keyspace so effects in **`dictFind`** (scoped to redis/redis#15320 key-length behavior) are measurable. The embedded description documents why concurrency and latency quantization make the sibling **pipeline-1 / single-connection** spec a poor instrument, and why **S vs P/R** and **`--key-maximum 11` with explicit `-n 11`** must stay as written.
> 
> **`dbconfig`** disables persistence, asserts **`keyspacelen: 11`**, and requests client resources (**10 CPUs**, 2g memory); **priority 69** and the usual **gcc/dockerhub** build variants are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd944d7bfa8ce27ea137ca187f86517e36e5f08a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Thanks — the client-thread point was the important one and I've applied it here, plus to the already-merged sibling in #564.

**Client sizing: agreed, changed.** Now `-c 5 -t 10` with `cpus: '10'`, which keeps the connection count exactly and matches what every other `_conns` spec in the suite does. Your framing is what convinced me: a clipped load generator caps both sides of a comparison equally, so the delta disappears and the run reads as "no regression" — the same false negative these specs exist to prevent, reached from the client side instead of via `-c 1`. Shipping a fix for one route while leaving another open would defeat the purpose.

fcostaoliveira:
On the two 562-specific points:

**Key pattern `R` → `P`: agreed, changed.** Your arithmetic is the argument — ~7.1 expected distinct keys per command out of 11 means roughly a third of the pairs rewrite a key already set earlier in the same command, the nine-10B/two-11B length mix only holds in expectation, and the varying distinct-lookup count adds spread to exactly the p50 this spec exists to resolve finely. `P` gives each connection `test_MSET1..test_MSET11` once per command, which is both the report's shape and the lower-variance option.

**`--key-maximum 11`: staying at 11, and I've documented why in the spec so it doesn't get "fixed" later.** You were right to flag it, and right that the `N+1` form is the tree-wide pattern — but every one of those preloads uses `-n allkeys`, which is where the count gets derived from the range and stops one short (#509). With an explicit `-n 11` nothing derives the count, and the sequential generator walks `key-minimum..key-maximum` inclusive.

The empirical check you asked for already exists, on the merged `memtier_benchmark-10keys-bitmap-50B-bitop-xor-pipeline-10`: it preloads `--key-minimum 1 --key-maximum 10 -n 10` against `check: keyspacelen: 10`, and it has been running on the fleet across both arches for months with real throughput numbers. Since `dbconfig_keyspacelen_check` runs before the client starts, a 9-key preload there would have failed every run rather than under-loading quietly. So the inclusive-with-explicit-`-n` behaviour is already demonstrated in production.

**On the p50 framing: fair, and softened.** You're right that at 500 outstanding requests p50 is largely queueing delay against a single-threaded server, so it tracks throughput more than per-command service time. The spec now says so explicitly and warns against reading it as interchangeable with the `-c 1` p50.

fcostaoliveira:
Coming back to the key-pattern question, and the Bugbot comment on the second commit was right — my earlier reply moving this to `P` was wrong, and I'd pushed it before that review landed. Switched to `S` in fd944d7.

The problem with `P` is the partition arithmetic. It slices the range per client as

```
range = (key_maximum - key_minimum) / total_clients + 1
min   = key_minimum + range * client_index
max   = min + range - 1          // last client: max = key_maximum
```

At `key_minimum 1`, `key_maximum 11`, `-c 5 -t 10` that's `range = 10/50 + 1 = 1` — one key per connection:

- connections 0–10 → keys 1..11 (the preloaded set)
- connections 11–48 → keys 12..49, **which the preload never wrote**
- connection 49 → `min = 50`, `max = 11`, an inverted range that just emits key 50 forever

So 39 of 50 connections miss the hot set entirely, and the main dict grows well past the 11 keys the whole spec is premised on. On top of that, each `__key__` placeholder in an arbitrary command draws its own index, so with a one-key slice every `MSET` collapses to the same key repeated 11 times. `keyspacelen: 11` wouldn't catch any of this, since it runs before the client starts.

`R` has the milder problem already noted — ~7.1 distinct keys per command in expectation, with the rest last-write-wins, and the 10B/11B name mix only holding on average.

`S` is the one that actually gives the reported shape. It never calls the per-client range split, so every connection keeps the full inclusive range and walks it with its own cursor; eleven consecutive placeholder draws are exactly `test_MSET1..test_MSET11`, one each, then it wraps. Eleven distinct keys every command, no per-command variance, and the nine-10B/two-11B split holds exactly rather than in expectation. The preload moves to `S` too so it no longer depends on its own client count.

On `--key-maximum 11`: staying at 11, and it's now written into the spec so it doesn't get "fixed" later. The two conventions in the tree don't actually conflict — the generator is inclusive of `key_maximum`, while `-n allkeys` derives its count as `key_maximum - key_minimum`, which is exactly why the `allkeys` specs carry the extra `+1` and this one, with an explicit `-n 11`, doesn't need it.

Also fixed the two stale `-c 25 -t 2` mentions in the description — leftovers from before the client topology change, thanks for catching those.

PR #563, opened 2026-09-11
test-suites: add a saturated HMSET listpack spec with the reported field names

fcostaoliveira:
Adds `memtier_benchmark-1key-hash-listpack-9-fields-short-values-hmset-pipeline-1-50_conns`.

### What it runs

`HMSET` overwrite of a single listpack-encoded hash with 9 short field/value pairs, at
`--pipeline 1` across 50 client connections (`-c 25 -t 2`). Metric of interest is **Ops/sec**.

### Why, when a 9-field listpack HMSET spec already exists

`...-hmset-pipeline-1` covers the right encoding and the right command, but it has two
fidelity gaps against the report it scopes (redis/redis#13876), and both of them cut exactly
the signal the issue is about:

**1. Connections.** The sibling is `-c 1 -t 1`. The report uses `redis-benchmark` with its
default `-c 50` and measures 280–290k ops/sec; the `-c 1` sibling measures around 40k. At one
connection the server is idle between requests, so round-trip time sets the throughput and a
per-command change of a few percent is diluted below run-to-run noise. Pipeline stays at 1
here because the report passes no `-P` — this changes the concurrency axis only.

**2. Field names.** The sibling uses `field4..field9` (6-byte names). The report uses
`HGET_HASH_FIELD1..HGET_HASH_FIELD6` (16-byte names). That is not cosmetic for this issue:
`lpFindCmp` compares field-name bytes at every traversal step, so 6-byte names exercise a
materially cheaper comparison than the names actually used in the report. This spec uses the
reported field names and the reported key name (`test_HMSET`).

I ran the `-c 1` / short-field-name sibling against this report and read its flat result as
"not reproduced." Both gaps above undercut that conclusion, so this spec exists to re-ask the
question with an instrument that can answer it.

### Encoding is unchanged

9 fields with 3–20 byte values stays under `hash-max-listpack-entries=128` and
`hash-max-listpack-value=64`, so the hash remains listpack-encoded throughout — same as the
sibling.

`cpus: '3'` for `-t 2` keeps the client clear of the saturation trap seen in #485/#486/#487.
The `-c 1` sibling is left in place for history continuity.

### Not yet run on the fleet

This spec has not run on the fleet yet. The preload issues the same single HMSET as the merged
sibling, and `check: keyspacelen: 1` will fail the run rather than report a wrong number.

---

> [!NOTE]
> **Low Risk**
> Adds a YAML benchmark definition only; no runtime, auth, or data-path code changes.
> 
> **Overview**
> Adds **`memtier_benchmark-1key-hash-listpack-9-fields-short-values-hmset-pipeline-1-50_conns`**, a fleet benchmark aimed at **redis/redis#13876** (HMSET / `lpFindCmp` regression). It reuses the same 9-field listpack HMSET workload as **`...-hmset-pipeline-1`**, but changes the run so it can surface server-side deltas the single-connection spec tends to miss.
> 
> The client runs **`--pipeline 1`** with **50 connections** (`-c 5 -t 10`, **`cpus: '10'`**) to mirror high-concurrency `redis-benchmark` reporting (~280–290k ops/sec) instead of ~40k at `-c 1 -t 1`. Commands use the reporter’s **`test_HMSET`** key and **16-byte `HGET_HASH_FIELD1..6`** names (plus the same username/age/location fields), replacing the sibling’s shorter `field4..field9` names and `test_hmset` key. Preload still seeds one key with **`keyspacelen: 1`**; metric is **Ops/sec** at priority **67**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53ca0e0ed888d2df78590861136369b3a0abff0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Thanks — the client-thread point was the important one and I've applied it here, plus to the already-merged sibling in #564.

**Client sizing: agreed, changed.** Now `-c 5 -t 10` with `cpus: '10'`, which keeps the connection count exactly and matches what every other `_conns` spec in the suite does. Your framing is what convinced me: a clipped load generator caps both sides of a comparison equally, so the delta disappears and the run reads as "no regression" — the same false negative these specs exist to prevent, reached from the client side instead of via `-c 1`. Shipping a fix for one route while leaving another open would defeat the purpose.

fcostaoliveira:
On the two 563-specific points:

**`lpFindCmp` wording: you're right, and I've tightened it.** It short-circuits on the encoded length (`slen == arg->slen`) and only reaches `memcmp` on a length match, so "compares field-name bytes on every traversal step" overstated the mechanism. The conclusion survives by a narrower route, which the spec now states: all six `HGET_HASH_FIELD*` names share a length, so the `memcmp`s that do run compare 16 bytes rather than 6, and the larger listpack means more bytes walked per lookup to reach the same field. Worth having that precise, given the whole premise is that name length isn't cosmetic.

**On `username`/`age`/`location` — those three are the report's too.** I can see why it looked inherited, since the sibling has them as well, but that's because the sibling was derived from the same report. The reporter's command is:

```
redis-benchmark -n 1000000 --csv HMSET test_HMSET \
  username alice age 30 location Paris \
  HGET_HASH_FIELD1 weugfuwebgfiwebviu HGET_HASH_FIELD2 ewuvbgevhberihvneirn \
  ...
```

So all nine pairs match the report, not six of nine. I've reworded the description to say that explicitly rather than leaving it ambiguous.

Good catch on `dataset_name` — that was deliberate for exactly the reason you identified. Reusing the sibling's label would have let `loaded_datasets` dedupe the preload away and left `test_HMSET` absent, since the key name and field names differ.

PR #569, opened 2026-09-16
test-suites: add a cache-resident saturated ZINCRBY listpack spec

fcostaoliveira:
Adds `memtier_benchmark-1key-zincrby-120-elements-50B-listpack-pipeline-10`.

### Why

We have `memtier_benchmark-1Mkeys-zincrby-120-elements-50B-listpack-pipeline-10`, which puts a 120-member / 50-byte listpack sorted set on each of 1,000,000 keys and drives `ZINCRBY` at one connection. That spec was added to chase a reported regression in the listpack existing-member update path, and it came back flat on both bisect endpoints. The flat result turned out to be a property of the spec rather than of the code: with a million keys, every `ZINCRBY` pays an object-plus-listpack cache miss just to reach the data, so the workload is DRAM-latency-bound and a difference of tens of nanoseconds inside `zsetAdd()`/`lpFind` is not resolvable on it in either direction.

This adds the opposite memory regime with the same data shape:

- **one key**, so the listpack stays cache-resident and the per-command cost is CPU-bound;
- **50 connections at pipeline 10** (`-t 2 -c 25`), so the server core saturates and `Ops/sec` is a metric the workload actually moves.

### What is held constant

The `ZADD` preload member/score list and the client `ZINCRBY` command are byte-identical to the 1Mkeys spec — same 120 members, same 50-byte member names, same integer scores, same `m60` target, same `zset-max-listpack-entries 128` / `zset-max-listpack-value 64`. Only the keyspace (1,000,000 → 1) and the concurrency (`-t 1 -c 1` → `-t 2 -c 25`) differ. Those two changes are the point of the spec, so it is a companion to the 1Mkeys one rather than a replacement, and the description says so explicitly.

### Shape notes

Two things a reviewer should check me on, both documented in the `description`:

1. The keyspace is exactly one key (`--key-minimum 1 --key-maximum 1`), so `--command-key-pattern=R` is degenerate — it can only select that key. It is spelled out rather than left to look like an unintentional uniform-over-one-key.
2. `ZINCRBY` grows `m60`'s score without bound, so within the first few dozen commands `m60` settles at the tail of the score-ordered listpack and stays there. The steady state is therefore a full ~120-member traversal per command — the longest traversal this shape can produce, which is what makes it sensitive to the path under test. The 1Mkeys spec has the same property, so the two remain comparable.

### Checklist

- `keyspacelen: 1` matches what the preload creates (`P` over an inclusive `[1,1]` with `-t 1 -c 1`).
- `tested-commands: [zincrby]` matches the only verb sent; preload `ZADD` is in the preload tool.
- Metric of interest named in the description (`Ops/sec`) and the load model saturates, so the two are consistent.
- `--hide-histogram` set, `--test-time` used for the measured phase, `-n allkeys` only in the preload, no `-x`.
- `-t 2` is within the client's 4-cpu request.

---

> [!NOTE]
> **Low Risk**
> Adds only a benchmark YAML definition; no runtime, auth, or data-path code changes.
> 
> **Overview**
> Adds a new memtier benchmark suite **`memtier_benchmark-1key-zincrby-120-elements-50B-listpack-pipeline-10`** as a **cache-resident, CPU-saturated** companion to the existing **1M-key** `ZINCRBY` listpack spec.
> 
> It keeps the same listpack shape and **`ZINCRBY m60`** workload but collapses the keyspace to **one** key (`keyspacelen: 1`) and drives load with **50 effective connections** (`-c 5 -t 10`, pipeline 10) so **Ops/sec** can reflect changes in the listpack existing-member update path. Preload uses a single **`ZADD`** with **`-n 1`** instead of **`allkeys`** on a one-key range.
> 
> The long **`description`** documents why this regime differs from the 1M-keys sibling (cache residency, saturation, and deeper steady-state listpack traversal on the single key).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2556119deb827a2d6487a0a3f9f3e2920e275b61. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Settled the preload bound question in 0dc736d, and the answer is that the two conventions in the tree don't conflict — they're answering different questions.

The generator is inclusive of `key_maximum`: the sequential path wraps on `next > m_key_max`, so a walk of `[1, N]` yields N keys. Separately, `-n allkeys` *derives* its request count as `key_maximum - key_minimum`. That subtraction is why the `allkeys` specs carry the extra `+1` — the 1Mkeys sibling needs `key_maximum 1000001` to get 1,000,000 requests — while a spec with an explicit `-n` doesn't, because nothing is being derived. Both in-tree patterns are correct for their own `-n`.

Which exposes a real bug here, though not the one you flagged. This preload had `--key-minimum 1 --key-maximum 1 -n allkeys`, so the derived count was `1 - 1 = 0`, and a zero count then falls back to the 10000-request default. It was replaying the same idempotent 120-member `ZADD` ten thousand times. The dataset it produced was correct, which is exactly why `keyspacelen: 1` never said anything — and it's also why the `1key-hash-hscan` spec you pointed at passes on the same shape. Now pinned to `-n 1`, which states the intent and keeps the preload O(1). Reasoning is in the spec description so it survives.

On the thread count — agreed, and changed. Now `-c 5 -t 10` with `cpus: '10'`, which keeps the documented 50 connections exactly while giving memtier enough threads to stay off the critical path. Your framing is the convincing one: a client-bound run caps both sides of the comparison equally, so the delta vanishes and it reads as "no regression" — the same false negative this spec exists to prevent, just reached from the client side. Same change I made on the MSET sibling in #562 for the same reason.

On the m60 steady-state point: fair, that sentence is doing more work than it can support. The claim holds for this spec — one key, so m60 reaches the tail almost immediately — but not for the 1Mkeys spec, where a million keys at `-c 1` means the average key never gets near the ~61 increments needed. I'll soften the "so the two remain comparable" framing; it doesn't affect this spec's own correctness, but it shouldn't be asserting something about the sibling that isn't true.

